PR #4, opened 2024-11-15
added initial layout for chrome extension

biw:
Hey @aidenybai, I saw that you already created a basis for the Chrome extension while I was doing this, so I tried to use as much existing code as I could. This PR includes a few improvements over the current:

- Setup using [Vite Plugin Web Extension](https://vite-plugin-web-extension.aklinker1.io/) to support easier hot reloading across content and background scripts
- Support for both Chrome/Firefox
- Shared types for message passing between background, popup, and content code
- Fixed CORES issues and got react-scan running on every page the user loads (via the [`Main` Execution World](https://developer.chrome.com/docs/extensions/reference/api/scripting#type-ExecutionWorld)
- Added the icon PNGs

It still needs a good amount of work as:
- The hot reloading isn't working on the popup
- There's no hooks to the react-scan code
- The react-scan code is inlined (to prevent CORES issues) but it's just the static build vs. caching it and [injecting it dynamically](https://developer.chrome.com/docs/extensions/develop/concepts/content-scripts#dynamic-declarative) as the unpkg script updates
- I didn't set up the bridge between the `Main` and `Isolated` content scripts, so while it runs on every page, it isn't sending events to the popup/background code.
- I tried to follow linting where I could, but there are a few linting errors 😅 
- I didn't want to force a monorepo setup, so there may be some pnpm config issues to solve

I don't want to be a blocker, so feel free to edit/merge/delete/etc, I just wanted to help out where I could. Thanks for writing and releasing `react-scan`!


aidenybai (approved):

aidenybai:
LFG thank you @biw

PR #16, opened 2024-11-18
docs(README.md): remove warning `Synchronous scripts should not be used.` for Next.js App router

matzar:
Remove warning `Synchronous scripts should not be used. See: https://nextjs.org/docs/messages/no-sync-scripts` for Next.js App router.

The docs recommend to not use synchronous scripts because I can impact webpage performance. 

I've checked with Next.js 14.2.1 and App router, and

```
<script src="https://unpkg.com/react-scan/dist/auto.global.js" async />
```

works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ export default function RootLayout({
   return (
     <html lang="en">
       <head>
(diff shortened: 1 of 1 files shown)


aidenybai:
thx! nice work

PR #28, opened 2024-11-22
Use custom element to encapsulate HTML elements

### LINKED ISSUE #27: Consider using Custom Element to encapsulate injected HTML elements
A common pattern that I've seen used in tools like Vercel's toolbar that are similar in spirit to React Scan is that they make use of [Shadow DOM](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_shadow_DOM) when injecting UI onto the page to encapsulate CSS and JS. Specifically, they create a custom element and attach all their UI to that custom element.

Right now I see that the toolbar and canvas overlay are appended in the <html> tag after the <body> element. What I'm proposing is that most of the code in the `start()` function is shifted into a custom element called `<react-scan-overlay>`.

Then the `start()` function likely becomes way simpler:

```ts
function start() {
  if (document.querySelector('react-scan-overlay')) return;
  const el = document.createElement('react-scan-overlay');
  document.body.appendChild(el);
}
```

This might be the one of the few use-cases where custom elements actually shine 😅, because you probably never want the document's styles or scripts to be able to access the toolbar or canvas overlay of React Scan, and shadow DOM gives you this encapsulation.

umar-ahmed:
- Resolves #27 
- Refactor the UI to use a custom element `<react-scan-overlay>` which encapsulates the canvas and toolbar styles


src/core/index.ts
```diff
@@ -153,14 +152,20 @@ export const setOptions = (options: Options) => {
 
 export const getOptions = () => ReactScanInternals.options;
 
-let inited = false;
+customElements.define('react-scan-overlay', ReactScanOverlay);
```
umar-ahmed, line 155: needs to be run somewhere at the global level to register the custom element

aidenybai:
@umar-ahmed nice work, thank you!

PR #29, opened 2024-11-22
Added Print Report function

davebcn87:
Adds printReport, where you could pass a list of component names to get specific rendering data.

## Example usage

Code: 

```jsx
useEffect(() => {
   printReport(['Button', 'AddTaskBar']);
}, [tasks, activeTab]);
```

Output: 

https://github.com/user-attachments/assets/6b237bb0-1ded-4e82-bd00-defffebc8a97

## Next 

I'd like to make this function available from console, so you could call it at any moment from the developer tools console and see actual stats.


davebcn87:
I'll close this one, as I don't think it makes more sense to use the console with all the tools you are adding in the modal screen

Issue #30, opened 2024-11-22
I suggest you clarify that it wont work in SSR context


bluematter:
```
  useEffect(() => {
    scan({
      enabled: process.env.NODE_ENV === 'development',
      log: true,
      includeChildren: true,
      report: true,
    });
  }, []);
```
  
  This within a main component works for me when using next.js app router
  
  Took me at least 30 minutes, when some documentation would have saved me time, and onboarded me faster.
  
  Nice work BTW

Armadillidiid:
Your comment just saved me minutes of debugging. Thanks.

aidenybai:
Closing — SSR context is now covered in the documentation.

Issue #26, opened 2024-11-22
Could we update the README for Remix project ?


theocerutti:
Could we update the README for Remix project ?

I tried to install it on my remix project but it doesn't work,

I installed the npm package and put it in the entry.client.tsx, I also tried the script tag

visormatt:
@theocerutti and @kryptowiz try to Install via npm and then update your `root.tsx` file with the following. Tried this approach after seeing the ticket here https://github.com/aidenybai/react-scan/issues/30. 

```tsx
  useEffect(() => {
    if (typeof document !== 'undefined') return;

    import('react-scan').then(({ scan }) => {
      scan({
        enabled: process.env.NODE_ENV === 'development',
        includeChildren: true,
        log: true,
        report: true,
      });
    });
  }, []);
```

theocerutti:
Yep working great

aidenybai:
Closing — workaround was confirmed working by the reporter. Remix setup info is available in the docs.

PR #31, opened 2024-11-23
export the "Options" interface

visormatt:
First, love the package 🤗  great stuff y'all. 

As I was getting this setup for some Remix apps in a monorepo I wanted to pull a hook out into a custom library. These would be useful to expose for TypeScript consumers (easy win). 

- https://github.com/aidenybai/react-scan/issues/30
- https://github.com/aidenybai/react-scan/issues/26

```tsx
import { useEffect } from 'react';
import type { Options } from 'react-scan';

export const useReactScan = (options: Options) => {
  useEffect(() => {
    if (typeof document === 'undefined') return;

    import('react-scan').then(({ scan }) => {
      scan(options);
    });
  }, []);
};
```

### DIFF
--- a/src/core/index.ts
+++ b/src/core/index.ts
@@ -13,7 +13,7 @@ import { createToolbar } from './web/toolbar';
 import { playGeigerClickSound } from './web/geiger';
 import { createPerfObserver } from './web/perf-observer';
 
-interface Options {
+export interface Options {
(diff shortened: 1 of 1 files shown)


aidenybai:
@visormatt nice work! do you also want to PR that hook, seems useful!

Issue #42, opened 2024-11-26
Why is the scrollbar being reset globally?


mthezi:
This library will globally change the appearance of scrollbars!
```css
* {
    scrollbar-width: thin;
    scrollbar-color: rgba(255, 255, 255, 0.3) rgba(255, 255, 255, 0.1);
}
```

I spent an hour troubleshooting the scrollbar issue on my page!

alexgarabt:
> Duplicate of [#42](https://github.com/aidenybai/react-scan/issues/42), I also been some time checking the styles on my project to see what it fails. I will try to make a pull request to fix the bug

I have made a [pull request](https://github.com/aidenybai/react-scan/pull/44) to fix the bug.

aidenybai:
Fixed in 0.0.32 @alexgarabt @mthezi

PR #44, opened 2024-11-26
Fix #41 issue and #42(duplicate of #41), scroll-bar global styles.

### LINKED ISSUE #41: Scrollbar issue
Hello,

When installing react-scan in my project it get rid of my scrollbar on my table (actually there are just not perceptible)

Before : 
Capture d’écran 2024-11-26 à 10 37 00

After:
Capture d’écran 2024-11-26 à 10 39 22

It looks like react scan add theses styles : 
Capture d’écran 2024-11-26 à 10 37 38

Could you make thoses to only target your code ?  and not make it globally (*)

Thanks a lot for your work :)

alexgarabt:
Fix the issue [#41](https://github.com/aidenybai/react-scan/issues/41) and duplicate [#42](https://github.com/aidenybai/react-scan/issues/42)  
Bug the styles of **react-scan** `scroll-bar` styles redefines normal styles of `scroll-bar` in projects that uses **react-scan**

Issue is due to define **react-scan** `scroll-bar` global styles in file `src/core/web/toolbar`, so the solution is to define those styles of the `scroll-bar` to target the local `<div>` with id `#react-scan-toolbar`  

You can see in the picture below that the the `scroll-bar` of my project is normal and the `scroll-bar` of **react-scan** still well.
solution_react_scan

### DIFF
--- a/src/core/web/toolbar.ts
+++ b/src/core/web/toolbar.ts
@@ -397,30 +397,30 @@ export const createToolbar = (): () => void => {
     background: rgba(255, 255, 255, 0.3);
   }
 
-  ::-webkit-scrollbar {
-  width: 4px;
-  height: 4px;
-}
-
-::-webkit-scrollbar-track {
-  background: rgba(255, 255, 255, 0.1);
-  border-radius: 4px;
-}
-
-::-webkit-scrollbar-thumb {
-  background: rgba(255, 255, 255, 0.3);
-  border-radius: 4px;
-}
-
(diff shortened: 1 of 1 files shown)


alexgarabt:
The issue starts in the commit [9c96b3b](https://github.com/aidenybai/react-scan/commit/9c96b3b58c7461ea5da6c647ab4497acff54ca87), when the styles that scopes global scrollbar are added

aidenybai:
nice work, thanks! @alexgarabt

PR #48, opened 2024-11-28
refactor(core): simplify toolbar bounds checking logic

xnmeet:
Problem
Quickly dragging the toolbar can move it outside the screen area.

Adjust

- Extract boundary calculations to improve readability
- Use position delta to determine updates

Now you can freely drag the toolbar without worrying about it going out of sight.


aidenybai:
nice work, thank you

PR #51, opened 2024-11-28
fix: resolve TypeScript issue with interval type

mnsdojo:
### Fix TypeScript Timer Type Issue

- Replaced `NodeJS.Timeout` with `NodeJS.Timer` to handle both `setTimeout()` and `setInterval()` correctly.
- This change resolves type errors and ensures compatibility with both one-time and recurring timers.


turaluix (approved):

aidenybai:
nice catch

PR #54, opened 2024-11-30
feat(core): add right resize handler

jiwon79:
### New Feature
Resolve the following issues to implement the right resize handler.

### Fixed Problems
|problem|as-is|to-be|
|--|--|--|
|If it goes outside the viewport during resizing, it can’t return to the viewport.|<video src="https://github.com/user-attachments/assets/e44cdfa3-7860-4618-b656-580694b4a806" />|<video src="https://github.com/user-attachments/assets/cc5bdf2e-849f-4436-91d2-7d5f70a71dce" />|
|If the toolbar’s width exceeds the inner width, there’s no way to make it smaller.|<video src="https://github.com/user-attachments/assets/7143eedc-672a-421e-993f-92072d60405f" />|<video src="https://github.com/user-attachments/assets/a860d96c-44b4-4288-ac79-1f395c3bf2f0" />|
||<video src="" />|<video src="" />|


pivanov:
I think this one is obsolete (at least the code base is changed)

PR #64, opened 2024-12-02
fix: add unique identifiers to tasks for accurate deletion

### LINKED ISSUE #61: Bug: Cannot delete task on web page
Preview:
zen_OWpFy4dUJK

Steps to replicate:
1. Open https://react-scan.com/
2. Add "1" task
3. Add "2" task
4. Add "1" task again!
5. Try to delete tasks

pradeepkaswan:
## Summary

This pull request addresses an issue where tasks with the same text could not be accurately deleted due to the lack of unique identifiers. By assigning a unique `id` to each task, we ensure that tasks can be correctly identified and managed.

## Changes Made

- **AddTaskBar Component**: Modified to assign a unique `id` to each task using `Date.now()`.
- **TaskList and TaskItem Components**: Updated to handle tasks with unique `id`s, ensuring accurate deletion.
- **State Management**: Updated the `tasks` state to be an array of objects, each containing an `id` and `text`.

## Why This Change?

Previously, tasks were identified by their text value, leading to conflicts when tasks with the same text were added. This change ensures that each task has a unique identifier, allowing for accurate management and deletion of tasks.

## How to Test

1. Add multiple tasks with the same text.
2. Delete tasks and verify that the correct task is removed.
3. Ensure that the task list updates correctly after deletion.

## Related Issue

Fixes #61

### DIFF
--- a/packages/website/src/index.jsx
+++ b/packages/website/src/index.jsx
@@ -79,13 +79,13 @@ export const App = () => {
               <AddTaskBar
                 onCreate={(value) => {
                   if (!value) return;
-                  setTasks([...tasks, value]);
+                  setTasks([...tasks, { id: Date.now(), text: value }]);
                 }}
               />
               <TaskList
                 tasks={tasks}
-                onDelete={(value) =>
-                  setTasks(tasks.filter((task) => task !== value))
+                onDelete={(id) =>
+                  setTasks(tasks.filter((task) => task.id !== id))
                 }
               />
             </div>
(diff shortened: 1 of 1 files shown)


pradeepkaswan:
> Hey @pradeepkaswan :) I just saw `Date.now()` and If accurate and reliable timing is critical, performance.now() is the better choice 🚀

 Thanks for the suggestion, @pivanov ! I have updated the timing function to use `performance.now()` for better accuracy. The changes have been pushed.

PR #65, opened 2024-12-02
fix: improve canvas clearing in inspect state machine

pivanov:
This change provides a more reliable solution for canvas clearing when transitioning between inspection states, preventing artifacts and improving performance.

### DIFF
--- a/packages/scan/src/core/web/inspect-element/inspect-state-machine.ts
+++ b/packages/scan/src/core/web/inspect-element/inspect-state-machine.ts
@@ -33,7 +33,7 @@ export type States =
 
 export const INSPECT_TOGGLE_ID = 'react-scan-inspect-element-toggle';
 export const INSPECT_OVERLAY_CANVAS_ID = 'react-scan-inspect-canvas';
-let lastHoveredElement: HTMLElement;
+let lastHoveredElement: HTMLElement | undefined
 let animationId: ReturnType<typeof requestAnimationFrame>;
 
 type Kinds = States['kind'];
@@ -98,17 +98,28 @@ export const createInspectElementStateMachine = () => {
   };
 
   const recursiveRaf = (cb: () => void) => {
+    let isActive = true;
+
     const helper = () => {
+      if (!isActive) return;
+
       if (animationId) {
         cancelAnimationFrame(animationId);
       }
 
       animationId = requestAnimationFrame(() => {
+        if (!isActive) return;
+
         cb();
         helper();
       });
     };
     helper();
+
+    return () => {
+      isActive = false;
+      cancelAnimationFrame(animationId);
+    };
   };
   ReactScanInternals.subscribeMultiple(
     ['reportDataByFiber', 'inspectState'],
@@ -121,8 +132,16 @@ export const createInspectElementStateMachine = () => {
             return;
           }
           case 'inspect-off': {
+            cancelAnimationFrame(animationId);
+
             clearCanvas();
-            return NO_OP;
+
+            unsubscribeAll();
+            lastHoveredElement = undefined;
+
+            return () => {
+              clearCanvas();
+            };
           }
           case 'inspecting': {
             recursiveRaf(() => {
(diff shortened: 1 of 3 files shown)


packages/scan/src/core/web/inspect-element/inspect-state-machine.ts
```diff
@@ -116,17 +122,7 @@ export const createInspectElementStateMachine = () => {
...
-            // i can't figure out why this happens, so this is an unfortunate hack
-            const mouseMove = () => {
-              clearCanvas();
-              updateCanvasSize(canvas, ctx);
-            };
-            window.addEventListener('mousemove', mouseMove);
-
-            return () => {
-              window.removeEventListener('mousemove', mouseMove);
-            };
```
RobPruzan, line 129: This was a weird case that I didn't do enough investigation on, so there is a root fix possible. But, the reproduction for the original problem that motivated this code was:

- move your mouse continuously while inspecting
- click escape (cancels the inspection)  and keep moving your mouse
- inspect overlay stays drawn on canvas

Testing this branch, the problem does come back. The best solution here would be figuring out which of the following cases is happening:
- the "inspecting" mouse move handler is not being cleaned up correctly, and drawing to screen during inspect-off
- some browser priority thing with how event handlers run in what order (queued mousemoves with stale closures?) 
  -  I don't think its this because we guard against that ` const mouseMove = throttle((e: MouseEvent) => {
              if (ReactScanInternals.inspectState.kind !== 'inspecting') {
                return;
              }`
 - we are clearing canvas wrong (something to do with how the canvas is transformed)
 - something else?

RobPruzan, line 129: This was a weird case that I didn't do enough investigation on, so there is a root fix possible. But, the reproduction for the original problem that motivated this code was:

- move your mouse continuously while inspecting
- click escape (cancels the inspection)  and keep moving your mouse
- inspect overlay stays drawn on canvas

Testing this branch, the problem does come back. The best solution here would be figuring out which of the following cases is happening:
- the "inspecting" mouse move handler is not being cleaned up correctly, and drawing to screen during inspect-off
- some browser priority thing with how event handlers run in what order (queued mousemoves with stale closures?) 
  -  I don't think its this because we guard against that ` const mouseMove = throttle((e: MouseEvent) => {
              if (ReactScanInternals.inspectState.kind !== 'inspecting') {
                return;
              }`
 - we are clearing canvas wrong (something to do with how the canvas is transformed)
 - something else?

RobPruzan (changes requested):
Here is the reproduction of the issue I mentioned in the review comment:
https://github.com/user-attachments/assets/2cb151c1-888e-47af-a643-f9dc4cd931aa

packages/scan/src/core/web/inspect-element/inspect-state-machine.ts
pivanov: Thank you for pointing this out 🚀

I’ve made the following changes to address the problem:
1. **Proper Cleanup of Animation Frames**:
   - Added an `isActive` flag in `recursiveRaf` to ensure that no stale animation frames continue running when transitioning to `inspect-off`.
   - Explicitly cancel the pending animation frame using `cancelAnimationFrame` when the state changes.

2. **Immediate Canvas Clearing**:
   - Ensured `clearCanvas()` is called immediately when transitioning to `inspect-off`, fixing the issue of the overlay remaining visible.

3. **Removed Unnecessary Hack**:
   - Removed the `mousemove`-based hack for clearing and resizing the canvas, as the root cause is addressed with proper cleanup.

4. **Event Handlers and State Cleanup**:
   - Used `unsubscribeAll()` to ensure all mouse-related event handlers and recursive RAF calls are unsubscribed when transitioning out of `inspecting` or `focused` states.
   - Reset `lastHoveredElement` to `undefined` to avoid stale references to DOM elements.

These changes should prevent the canvas overlay from persisting after `inspect-off` is triggered, ensuring proper state transitions.

Let me know if there’s anything I missed or further improvements you’d suggest!

pivanov:
This one is obsolate now :)

PR #66, opened 2024-12-02
perf: optimize overlay rendering and state handling

pivanov:
This optimization focuses on improving the element inspection overlay performance and code clarity. Key changes:

1. Introduces element caching to prevent redundant redraws
2. Consolidates overlay update logic with a unified type parameter
3. Removes redundant state variables (currentHoveredElement)
4. Improves event handling order and code organization
5. Maintains proper state transitions while reducing unnecessary operations

The changes make the inspection tool more efficient and the code more maintainable.

https://github.com/user-attachments/assets/de394826-f44f-4c72-b5eb-0aa490019de9

### DIFF
--- a/packages/scan/src/core/web/inspect-element/inspect-state-machine.ts
+++ b/packages/scan/src/core/web/inspect-element/inspect-state-machine.ts
@@ -32,7 +32,6 @@ export type States =
 
 export const INSPECT_TOGGLE_ID = 'react-scan-inspect-element-toggle';
 export const INSPECT_OVERLAY_CANVAS_ID = 'react-scan-inspect-canvas';
-let lastHoveredElement: HTMLElement;
 let animationId: ReturnType<typeof requestAnimationFrame>;
 
 type Kinds = States['kind'];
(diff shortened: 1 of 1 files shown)


aidenybai:
wow this is fast asf

pivanov:
> wow this is fast asf
Yep :) I have few more ideas but I will wait to merge (or improve this one if needed) and I will bring the rest ;)

packages/scan/src/core/web/inspect-element/inspect-state-machine.ts
```diff
@@ -403,7 +427,6 @@ export const createInspectElementStateMachine = () => {
             };
           }
         }
-        inspectState satisfies never;
```
RobPruzan, line 406: This isn't dead code, it confirms the switch/case handling is exhaustive. Typescript will throw an error if a case is added to the discriminated union and the case is not handled here

RobPruzan (changes requested):
Couple comments left

Also, this feature itself is super fragile, ill make an effort to document edge cases more, but some websites to try out the changes to make sure nothing broke:

https://x.com/home
https://github.com/aidenybai/react-scan (this page specifically)
https://railway.app/
https://replit.com/

How i test it is spam my mouse around, click around, make sure it all works (not very scientific :p) 

A good workflow to test on any website is:
1. run `pnpm build`
2. go to packages/scan/dist/auto.global.js and copy the entire script
3. paste it into console of website


Also I saw a couple times my cpu hit 100% when toggling the inspect element button when an element is selected. I'm not confident it came from this branch (encountered it on this page [railway](https://railway.app/) and [this github page](https://github.com/aidenybai/react-scan/pull/66/files#) , but I saw it twice so wanted to point it out incase you see it too (couldn't reproduce on main)

packages/scan/src/core/web/inspect-element/inspect-state-machine.ts
pivanov: oh ... format on auto save 😮‍💨 
I will make it like:
```
default: {
  inspectState satisfies never;
  return;
}
```

RobPruzan: lol thats brutal. but i guess it makes sense it looks like deadcode to static analysis

it's fine as is tbh, already validated at type level it's exhaustive

pivanov:
I changed from tracking window-level metrics (scroll position, window size) to directly comparing element positions using `getBoundingClientRect()`. Here's why I think this is a better approach:

#### More accurate position tracking
- `getBoundingClientRect()` gives me the element's position relative to the viewport
- Catches all types of position changes (scrolling, transforms, resizing)
- Works correctly with scrollable containers at any level of nesting

#### More reliable
- Window metrics could miss position changes from container scrolling
- Previous approach wouldn't catch CSS transforms
- Element's actual position is more important than what caused the change

#### Simpler code
- I removed multiple tracking variables
- I removed separate scroll event listeners
- I now have a single source of truth for element position

## Better edge case handling
- Works with nested scrollable containers
- Handles CSS transforms properly
- More accurate for complex layouts

<br />

**Tradeoff**: Slightly more frequent `getBoundingClientRect()` calls, but since I'm already in a RAF loop and only checking one element, the performance impact is minimal compared to the benefits of more reliable position tracking.

<br />

https://github.com/user-attachments/assets/2bdd0285-3183-4f3c-bdb8-994b3f6f1043

packages/scan/src/core/web/inspect-element/inspect-state-machine.ts
pivanov: Sure, I'll leave it as is.

pivanov:
This one is obsolate now :)

PR #80, opened 2024-12-06
transpile to es2018

mxmul:
I've been using `react-scan` in a project with a legacy webpack 4 build, and noticed that the NPM dist uses modern syntax that can't be parsed. To work around, I've been running `react-scan` through babel-loader, but this PR should fix the problem.

It seems that webpack 4 uses acorn@6 to parse JS, and that it only supports es2018 syntax[^1]. Changing the compilation target for `esnext` to `es2018` fixes the issue, and doesn't affect the build size too much (52kb -> 54kb).

Tested with a trivial webpack 4 config:

| Before | After |
| -- | -- |
| Screenshot 2024-12-05 at 6 58 18 PM | Screenshot 2024-12-05 at 6 58 57 PM |

[^1]: https://github.com/acornjs/acorn/blob/6.x/acorn/README.md

### DIFF
--- a/packages/scan/package.json
+++ b/packages/scan/package.json
@@ -96,7 +96,7 @@
   },
   "main": "dist/index.js",
   "module": "dist/index.mjs",
-  "browser": "dist/auto.global.js",
+  "browser": "dist/index.js",
   "types": "dist/index.d.ts",
   "bin": "bin/cli.js",
   "files": [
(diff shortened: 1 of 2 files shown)


packages/scan/package.json
```diff
@@ -96,7 +96,7 @@
   },
   "main": "dist/index.js",
   "module": "dist/index.mjs",
-  "browser": "dist/auto.global.js",
+  "browser": "dist/index.js",
```
mxmul, line 99: webpack prefers the `browser` field by default. When bundling, I think consumers will generally want the side-effect free module.

PR #99, opened 2024-12-10
feat(scan): add Astro monitoring component

xinyao27:
Add monitoring support for astro, usage:

```
// Layout.astro
---
import Monitoring from 'react-scan/monitoring/astro'
---

<Monitoring apiKey="" url="https://monitoring.react-scan.com/api/v1/ingest"/>
```

### DIFF
--- a/packages/scan/package.json
+++ b/packages/scan/package.json
@@ -173,6 +173,9 @@
       "monitoring/remix": [
         "./dist/core/monitor/params/remix.d.ts"
       ],
+      "monitoring/astro": [
+        "./dist/core/monitor/params/astro/index.ts"
+      ],
       "react-component-name/vite": [
         "./dist/react-component-name/vite.d.ts"
       ],
--- a/packages/scan/package.json
+++ b/packages/scan/package.json
@@ -51,6 +51,9 @@
       "import": "./dist/core/monitor/params/remix.mjs",
       "require": "./dist/core/monitor/params/remix.js"
     },
+    "./monitoring/astro": {
+      "import": "./dist/core/monitor/params/astro/index.ts"
+    },
     ".": {
       "production": {
         "import": {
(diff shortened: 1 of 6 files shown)


packages/scan/package.json
```diff
@@ -200,9 +203,11 @@
     "auto.d.ts"
   ],
   "scripts": {
+    "copy-astro": "cp -R src/core/monitor/params/astro dist/core/monitor/params",
```
RobPruzan, line 206: why is this needed? Seems astro uses vite, so if there's nothing special that I'm not aware of, should be able to follow the same implementation as `react-scan/monitoring/react-router`

packages/scan/package.json
```diff
@@ -51,6 +51,9 @@
       "import": "./dist/core/monitor/params/remix.mjs",
       "require": "./dist/core/monitor/params/remix.js"
     },
```
RobPruzan, line 53: Make sure to update the "typesVersions" too. It's needed for older versions of typescript that doesn't support the package.json exports field

RobPruzan (changes requested):
Okay tested it and it works :) 

Last change I didn't already mention in the comments, can we make the Monitoring export from /monitor/astro a named export instead of default export just to stay consistent with the other framework exports

packages/scan/package.json
xinyao27: packages/scan/src/core/monitor/params/astro/Monitoring.astro
packages/scan/src/core/monitor/params/astro/index.ts

The above two files will not be packaged, they will be exported directly and parsed by astro.
So these two files need to be manually copied into the dist directory.

xinyao27:
> Okay tested it and it works :)
> 
> Last change I didn't already mention in the comments, can we make the Monitoring export from /monitor/astro a named export instead of default export just to stay consistent with the other framework exports

Updated, new way:

```
// Layout.astro
---
import { Monitoring } from 'react-scan/monitoring/astro'
---

<Monitoring apiKey="" url="https://monitoring.react-scan.com/api/v1/ingest"/>
```

RobPruzan (approved):
lgtm

PR #101, opened 2024-12-11
feat(scan): allow users to have toolbar mounted while enabled as false

Looskie:
Previously, there was no way to mount the toolbar while having the scanning turned off. Now, providing the config below will allow users to have scanning turned off by default, but still have the toolbar mounted, acting as an opt-in with the toolbar.

```typescript
scan({
  enabled: false,
  showToolbar: true
});
```


aidenybai:
ty!!

PR #106, opened 2024-12-12
fix: SDK irregularities and commits / branch addition

NisargIO:


packages/scan/src/core/monitor/network.ts
```diff
@@ -6,19 +6,29 @@ import {
...
+    return;
+  }
   const now = performance.now();
-  const pendingInteractions: typeof interactions = [];
-  const completedInteractions: typeof interactions = [];
-
-  interactions.forEach((interaction) => {
+  // We might trigger flush before the interaction is completed,
+  // so we need to split them into pending and completed by an arbitrary time.
+  const pendingInteractions = new Array<InternalInteraction>();
```
aidenybai, line 26: no

aidenybai (changes requested):
no

packages/scan/src/core/monitor/network.ts
NisargIO: it's faster 
image

PR #107, opened 2024-12-12
fix(scan): fix promise/proxy handling and optimize object inspection

pivanov:
Improving the handling of proxy objects and promises and ensures that complex objects, like `Next.js` params or URLSearchParams, are inspected accurately without errors or performance issues.

Key Changes:
- proxy detection and unwrapping
- Improved handling of Promise and Symbol.iterator edge cases.
- optimized object inspection by consolidating operations for better performance.

These updates fix bugs in the property inspector and ensure a smoother, error-free experience when dealing with complex object types.

Result:
image

Thanks for the guidance of @RobPruzan 🙇‍♂️

### DIFF
--- a/packages/scan/src/core/web/inspect-element/view-state.ts
+++ b/packages/scan/src/core/web/inspect-element/view-state.ts
@@ -17,6 +17,45 @@ export const cumulativeChanges = {
   context: new Map<string, number>(),
 };
 
+const getProxyValue = (proxy: any) => {
+  try {
+    if (!proxy || typeof proxy !== 'object') {
+      return proxy;
+    }
+
+    // Handle URLSearchParams-like objects
+    if (typeof proxy.entries === 'function' && typeof proxy.get === 'function') {
+      try {
+        const entries = Array.from(proxy.entries() as Iterable<[string, any]>);
+        return Object.fromEntries(entries);
+      } catch (err) {
+        return proxy;
+      }
+    }
+
+    // Handle regular objects more efficiently
+    try {
+      const descriptors = Object.getOwnPropertyDescriptors(proxy);
+      const result: Record<string, any> = {};
+
+      // Using for...in is faster than reduce for this case
+      for (const key in descriptors) {
+        const descriptor = descriptors[key];
+        if (typeof key === 'string' && 'value' in descriptor) {
+          result[key] = descriptor.value;
+        }
+      }
+
+      return result;
+    } catch (err) {
+      return proxy;
+    }
+
+  } catch (err) {
+    return proxy;
+  }
+};
+
 export const renderPropsAndState = (didRender: boolean, fiber: any) => {
   const propContainer = Store.inspectState.value.propContainer;
 
@@ -380,6 +419,7 @@ export const createPropertyElement = (
     if (isExpandable) {
       const isExpanded = EXPANDED_PATHS.has(currentPath);
 
+      // Check for circular references first
       if (typeof value === 'object' && value !== null) {
         let paths = objectPathMap.get(value);
         if (!paths) {
@@ -392,6 +432,28 @@ export const createPropertyElement = (
         paths.add(currentPath);
       }
 
+      const unwrapped = getProxyValue(value);
+      const isNonExpandable = (unwrapped === value &&
+        value &&
+        Object.getPrototypeOf(value)?.constructor?.name === 'Proxy') ||
+        value instanceof Promise;
+
+      // For non-expandable items, render like a simple property
+      if (isNonExpandable) {
+        const preview = document.createElement('div');
+        preview.className = 'react-scan-preview-line';
+        preview.dataset.key = key;
+        preview.dataset.section = section;
+        preview.innerHTML = `
+          <span style="width: 8px; display: inline-block"></span>
+          <span class="react-scan-key">${key}:&nbsp;</span>
+          <span class="${getValueClassName(value)}">${getValuePreview(value)}</span>
+        `;
+        container.appendChild(preview);
+        return container;
+      }
+
+      // Normal expandable logic for other objects
       container.classList.add('react-scan-expandable');
       if (isExpanded) {
         container.classList.add('react-scan-expanded');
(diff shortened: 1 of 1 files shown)


packages/scan/src/core/web/inspect-element/view-state.ts
```diff
@@ -737,14 +811,35 @@ export const getValuePreview = (value: any) => {
...
+        if (value === null) return 'null';
+        if (value === undefined) return 'undefined';
+
+        // Handle special built-in types first
+        if (value instanceof Promise) return '[Promise]';
+        if (value instanceof Set) return `Set(${value.size})`;
+        if (value instanceof Map) return `Map(${value.size})`;
+
+        // Try to unwrap proxy values
+        const proto = Object.getPrototypeOf(value);
```
RobPruzan, line 824: Does this work? in node repl:

```
> const proxy = new Proxy(target, handler);
>
> proxy
Proxy [
  { message: 'Hello, world!' },
  { get: [Function: get], set: [Function: set] }
]
> const proto = Object.getPrototypeOf(proxy);
> proto
[Object: null prototype] {}
> proto?.constructor?.name
'Object'
>
```

packages/scan/src/core/web/inspect-element/view-state.ts
```diff
@@ -12,6 +12,55 @@ const EXPANDED_PATHS = new Set<string>();
...
+
+
+const getProxyValue = (proxy: any) => {
+  try {
+    if (!proxy || typeof proxy !== 'object') {
+      return proxy
+    };
+
+    // Handle URLSearchParams-like objects
+    if (proxy[Symbol.iterator]) {
```
RobPruzan, line 45: Wouldn't this still cause the same next.js error? Iterating over the search params will cause next to show an error to user, even if you catch the error

packages/scan/src/core/web/inspect-element/view-state.ts
lxsmnsyc: yeah this doesn't work normally, neither is `instanceof`. We probably shouldn't need to distinguish proxies.

packages/scan/src/core/web/inspect-element/view-state.ts
```diff
@@ -12,6 +12,55 @@ const EXPANDED_PATHS = new Set<string>();
 const fadeOutTimers = new WeakMap<HTMLElement, ReturnType<typeof setTimeout>>();
 const disabledButtons = new Set<HTMLButtonElement>();
 
+// Utility to check and unwrap proxies
+const isProxy = (obj: any) => Object.getPrototypeOf(obj)?.constructor?.name === 'Proxy';
```
lxsmnsyc, line 16: This doesn't work atm.

packages/scan/src/core/web/inspect-element/view-state.ts
pivanov: You're absolutely right ... the proxy detection was unreliable. I've removed the isProxy check and refactored the code to focus on safely accessing properties instead of trying to detect proxies.
Much cleaner approach, thank you for catching this!

packages/scan/src/core/web/inspect-element/view-state.ts
pivanov: Good catch again ...  I've updated the code to use `entries()` and `get()` method checks instead of `Symbol.iterator`.
This should handle `URLSearchParams` like objects more safely without triggering Next.js errors.

packages/scan/src/core/web/inspect-element/view-state.ts
pivanov: Completely agree! Trying to distinguish proxies and using `instanceof` was the wrong approach.
I've refactored the code to use capability checks instead (checking for specific methods like then, add, set, etc). This is much more reliable and follows the duck typing principle. Thanks for the guidance! 🚀

packages/scan/src/core/web/inspect-element/view-state.ts
```diff
@@ -12,6 +12,50 @@ const EXPANDED_PATHS = new Set<string>();
...
+        // If conversion fails, return original
+        return proxy;
+      }
+    }
+
+    // Instead of checking for proxy, try to safely get own properties
+    try {
+      const descriptors = Object.getOwnPropertyDescriptors(proxy);
+      const unwrapped = Object.fromEntries(
+        Object.entries(descriptors).reduce<Array<[string, any]>>((acc, [key, descriptor]) => {
```
lxsmnsyc, line 39: I'll allow this for now, but please mark this as a "FIXME" for perf improvements. `Array.prototype.reduce` is slow.

packages/scan/src/core/web/inspect-element/view-state.ts
```diff
@@ -439,6 +484,28 @@ export const createPropertyElement = (
...
 
+      const unwrapped = getProxyValue(value);
+      const isNonExpandable = (unwrapped === value &&
+        value &&
+        Object.getPrototypeOf(value)?.constructor?.name === 'Proxy') ||
+        value instanceof Promise;
+
+      // For non-expandable items, render like a simple property
+      if (isNonExpandable) {
+        const preview = document.createElement('div');
```
lxsmnsyc, line 495: I'm legit thinking if `cloneNode` is possible here. It's a lot faster and easier to setup as a template.

packages/scan/src/core/web/inspect-element/view-state.ts
```diff
@@ -519,77 +586,79 @@ export const createPropertyElement = (
...
+                    new WeakMap(),
+                  );
+                  if (!el) {
+                    return;
+                  }
+                  arrayContainer.appendChild(el);
+                });
+                content.appendChild(arrayContainer);
+              } else {
+                Object.entries(value).forEach(([k, v]) => {
```
lxsmnsyc, line 626: If it's possible, please use `for of` over forEach everytime.

packages/scan/src/core/web/inspect-element/view-state.ts
```diff
@@ -724,30 +793,51 @@ export const getValueClassName = (value: any) => {
...
+
+    switch (typeof value) {
+      case 'string':
+        return `"${value}"`;
+      case 'number':
+      case 'boolean':
+        return String(value);
+      case 'object': {
+        // Safe checks for common interfaces rather than type detection
+        if (value.then && typeof value.then === 'function') {
```
lxsmnsyc, line 812: hmmm, not sure if `'then' in value` is faster.

packages/scan/src/core/web/inspect-element/view-state.ts
pivanov: I will start working to migrate all of this to react/preact

packages/scan/src/core/web/inspect-element/view-state.ts
pivanov: 👍  yeah it's just old code wrapped in `if`

packages/scan/src/core/web/inspect-element/view-state.ts
pivanov: I will fix it ... you are right!

pivanov:
obsolate

PR #115, opened 2024-12-13
perf: offscreen canvas for outlines

lxsmnsyc:
- [x] Adds `SmolWorker`, a utility class for creating workers.
- [x] Move the drawing part of outlines to offscreen.


packages/scan/src/core/web/overlay.ts
```diff
@@ -22,39 +23,62 @@ export const initReactScanOverlay = () => {
...
-      });
+      updateCanvasSize();
     }
   });
 
   window.addEventListener('scroll', () => {
     recalcOutlines();
   });
 
+  const offscreen = overlayElement.transferControlToOffscreen();
```
RobPruzan, line 67: only thing that's maybe of concern is this landed support in safari on 2023-03-27, so pretty recent. 

Not sure what a correct heuristic for something like this. I can't remember the last time i upgraded safari, but i do use it once in a while, so if there's an alternative/fallback that would be ideal

lxsmnsyc: yeah I've been thinking back-and-forth about this one. Fallback is simple, but then I remembered using `OffscreenCanvas` directly for `measureTextCached` (since overlay canvas context is no longer accessible on the main thread)

I need a polyfill that works in place for OffscreenCanvas

lxsmnsyc: also the OffscreenCanvas API is mainly at 95%, the lowest safe percentage

RobPruzan: the most trivial thing we could do is draw to normal canvas if not available

lxsmnsyc: yeah that's like the fallback solution I've seen the most.

PR #128, opened 2024-12-17
fix(scan-core): enabling the sound toggle should not close the toolbar.

theborowski:
## context

The audio toggle was causing the toolbar to be closed since there was check that was returning true whenever `options.showToolbar` was falsy. 

## changes
* explicitly check `options.showToolbar === false` when determining if the toolbar should be removed.
* update the docs for testing
* add a comment to the `playGeigerClickSound` function about why audio does not work in Firefox

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -17,8 +17,8 @@ pnpm run dev
 ## 3. Run test environment
 
 ```bash
-cd test
+cd packages/kitchen-sink
 pnpm run dev
 ```
 
-Go to http://localhost:5173 to see the test environment.
+Go to http://localhost:5175 to see the test environment.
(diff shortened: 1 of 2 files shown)


pivanov:
Hey @theborowski thanks for this one
I just make some geiger improvements in #133

can you touch your PR a bit?

Also about this [line](https://github.com/aidenybai/react-scan/pull/128/files#diff-8403cbc616fb49b70b44b1cee8c52f5d4926a98368a8f7b36353d8cec5e728daR230) I think @aidenybai can say is this the right behaviour

theborowski:
Thanks @pivanov, I just rebased this branch against your changes. I appreciate the Firefox fix as I was tinkering in there earlier but couldn't quite get the geiger to sound right in Firefox.

theborowski:
@pivanov looks like you've already fixed this here https://github.com/aidenybai/react-scan/commit/c9efa9f2ecfc39063a635b8c9d8f752363d98ab3 so I'll close this one out.

PR #131, opened 2024-12-17
fix: nextjs issue with requestAnimationFrame

pivanov:


pivanov:
Closed in favor of #130

PR #136, opened 2024-12-17
chore(kitchen-sink): update dependencies and fix a key issue in the todo list

theborowski:
## changes
* bump `react` and `react-dom` from the RC version to the official `19.0.0` release
* remove unused `babel-plugin-react-compiler` dependency
* move `vite` related dependencies into `devDependencies`
* fix a tiny bug where adding two of the same text items in the list would result in a key error 
* prevent adding an item to the list with the enter key if it's text is blank

## visual diff

https://github.com/user-attachments/assets/9ccb203b-cc12-43bb-bf69-efcf013689d6

### DIFF
--- a/packages/kitchen-sink/src/index.jsx
+++ b/packages/kitchen-sink/src/index.jsx
@@ -60,8 +60,8 @@ export const App = () => {
 export const TaskList = ({ tasks, onDelete }) => {
   return (
     <ul>
-      {tasks.map((task) => (
-        <TaskItem key={task} task={task} onDelete={onDelete} />
+      {tasks.map((task, i) => (
+        <TaskItem key={`${task}_${i}`} task={task} onDelete={onDelete} />
       ))}
     </ul>
   );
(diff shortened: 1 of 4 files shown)


packages/kitchen-sink/src/index.jsx
```diff
@@ -59,8 +59,8 @@ export const App = () => {
 export const TaskList = ({ tasks, onDelete }) => {
   return (
     <ul>
-      {tasks.map((task) => (
-        <TaskItem key={task} task={task} onDelete={onDelete} />
+      {tasks.map((task, i) => (
+        <TaskItem key={`${task}_${i}`} task={task} onDelete={onDelete} />
```
lxsmnsyc, line 63: honestly this is the same even if we remove the `key` prop. Needs an alternative "id generation"

theborowski: Do you think just using a timestamp here as an "id" is sufficient?

lxsmnsyc: we need something like

```ts
interface Task {
  id: string;
  text: string;
}
```

and then use `id` as the key.

theborowski: I refactored this to use an object as you suggested. This fixes the `key` problem as well as improves deletion with using the `id` of the task. I also moved the task creation in a function so that clicking and pressing enter have the same behavior.

theborowski:
@lxsmnsyc are there any additional changes needed in order to get this merged in?

lxsmnsyc:
Nothing specifically. LGTM

PR #137, opened 2024-12-17
fix(docs): e.g. → i.e.

devgru:
Looks like server components are not an example here


lxsmnsyc:
I think `e.g.` works fine since there are many cases it can occur "outside of a `use server`"

PR #138, opened 2024-12-17
chore(scan): optimize perf of core scan

RobPruzan:
Summary:
- use IntersectionObserver over getClientBoundingRect to minimize reflows
- optimize fast serialize
- sample renders to run isRenderUnncessaryOn
- dont merge outlines when there are significant active outlines
- flush async instead of in hot path/ avoid promise creation every render
- dont draw outlines that are out of viewport

### Before ( ~1600ms react scan blocking time):

image

### After (~170ms react scan blocking time)
image

### DIFF
--- a/packages/scan/src/core/instrumentation.ts
+++ b/packages/scan/src/core/instrumentation.ts
@@ -92,59 +92,72 @@ export interface Render {
   count: number;
   forget: boolean;
   changes: Array<Change> | null;
-  unnecessary: boolean;
+  unnecessary: boolean | null;
   didCommit: boolean;
   fps: number;
 }
 
 const unstableTypes = ['function', 'object'];
 
-export const fastSerialize = (value: unknown, depth = 0) => {
+const cache = new WeakMap<object, string>();
+
+export function fastSerialize(value: unknown, depth = 0): string {
   if (depth < 0) return '…';
+
   switch (typeof value) {
     case 'function':
       return value.toString();
     case 'string':
       return value;
+    case 'number':
+    case 'boolean':
+    case 'undefined':
+      return String(value);
     case 'object':
-      if (value === null) {
-        return 'null';
-      }
-      if (Array.isArray(value)) {
-        return value.length > 0 ? `[${value.length}]` : '[]';
-      }
-      if (isValidElement(value)) {
-        // attempt to extract some name from the component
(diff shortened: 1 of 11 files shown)


packages/scan/src/core/instrumentation.ts
```diff
@@ -92,59 +92,72 @@ export interface Render {
...
-export const fastSerialize = (value: unknown, depth = 0) => {
+const cache = new WeakMap<object, string>();
+
+export function fastSerialize(value: unknown, depth = 0): string {
   if (depth < 0) return '…';
+
   switch (typeof value) {
     case 'function':
-      return value.toString();
+      return 'function'; // function.toString can be slow
```
aidenybai, line 109: can't do stuff like this since we need the value to determine if it's unstable

packages/scan/src/core/instrumentation.ts
```diff
@@ -92,59 +92,72 @@ export interface Render {
...
     case 'string':
       return value;
+    case 'number':
+    case 'boolean':
+    case 'undefined':
+      return String(value);
     case 'object':
```
aidenybai, line 116: can't do this as well since we need a simple way to check unstable objects

RobPruzan: elaborate on this pls

packages/scan/src/core/web/utils/outline.ts
```diff
@@ -16,8 +13,8 @@ export interface ActiveOutline {
   alpha: number;
   frame: number;
   totalFrames: number;
-  resolve: () => void;
```
lxsmnsyc, line 19: I had a cleanup for this at #129 but this is fine

RobPruzan: if yours ready to merge? Ill rebase ontop if thats the case

packages/scan/src/core/instrumentation.ts
aidenybai: say 2 function values (given via toString) are same but reference is not, this means it needs to be optimized

RobPruzan: yeah fixed that comment, referring to this
> can't do this as well since we need a simple way to check unstable objects

aidenybai: gud to go

PR #145, opened 2024-12-18
feat: add persistent options with localStorage

pivanov:


packages/scan/src/core/index.ts
```diff
@@ -225,32 +295,57 @@ export const getReport = (type?: React.ComponentType<any>) => {
   return Store.legacyReportData;
 };
 
-export const setOptions = (options: Options) => {
+const initializeScanOptions = (userOptions?: Partial<Options>) => {
+  const options = ReactScanInternals.options.value;
+
+  const localstorageOptions = readLocalStorage<LocalStorageOptions>('react-scan-options');
```
aidenybai, line 301: localStorage, not localstorage

PR #149, opened 2024-12-19
Memory leak

RobPruzan:
Fixes memory leaked caused by frequent accumulation of re-renders in a large array

- aggregates renders together in scheduled/active outlines immediately, instead of doing it when painting.
  - this required a large rewrite of painting logic
- re-use objects when possible
- remove mergeOutline merge limit

## Before

https://github.com/user-attachments/assets/748a228e-9ea3-45fc-b93f-2991310180c9

https://github.com/user-attachments/assets/f001647d-e60e-4352-baeb-daa32003df6c


packages/scan/src/core/index.ts
```diff
@@ -340,7 +346,7 @@ export const reportRender = (fiber: Fiber, renders: Array<Render>) => {
   const { selfTime } = getTimings(fiber.type);
   const displayName = getDisplayName(fiber.type);
 
-  Store.lastReportTime.value = performance.now();
+  Store.lastReportTime.value = Date.now();
```
RobPruzan, line 349: performance.now() is slow and we don't need the precision

packages/scan/src/core/instrumentation.ts
```diff
@@ -75,23 +75,29 @@ export const isElementInViewport = (
...
   type: 'props' | 'context' | 'state';
   name: string;
   prevValue: unknown;
   nextValue: unknown;
   unstable: boolean;
 }
 
+// this must grow O(1) w.r.t to renders
+export interface AggregatedChange {
+  type: Set<'props' | 'context' | 'state'>;
```
lxsmnsyc, line 88: hmmm this in the form of bit flags might be possible here

packages/scan/src/core/utils.ts
```diff
@@ -53,6 +85,7 @@ export const getLabelText = (renders: Array<Render>) => {
   const parts: Array<string> = [];
   let cumulativeTime = 0;
   for (const count of sortedCounts) {
+    // biome-ignore lint/style/noNonNullAssertion: <explanation>
```
lxsmnsyc, line 88: weird, the linter should've been disabled in main

RobPruzan: oh that's probably old before i rebased

RobPruzan: yeah it is, removed them

packages/scan/src/core/monitor/debug.js
```diff
@@ -0,0 +1,131 @@
+(function () {
+  function getInteractionType(eventName) {
+    if (['pointerdown', 'pointerup', 'click'].includes(eventName)) {
```
lxsmnsyc, line 3: switch might be better in this whole function

packages/scan/src/core/web/utils/helpers.ts
```diff
@@ -63,7 +60,7 @@ export const debounce = <T extends (...args: Array<any>) => any>(
       timeoutId = window.setTimeout(() => {
         isLeadingInvoked = false;
         timeoutId = undefined;
-        fn(...(lastArgs!));
+        fn(...lastArgs!);
```
lxsmnsyc, line 63: `fn.apply(null, lastArgs)`, but the linter might bork

packages/scan/src/core/monitor/debug.js
RobPruzan: oops my b that was just to debug something i didn't mean to commit it

packages/scan/src/core/web/utils/helpers.ts
```diff
@@ -129,6 +112,9 @@ export const saveLocalStorage = <T>(storageKey: string, state: T): | void => {
   }
 };
 
-export const toggleMultipleClasses = (element: HTMLElement, ...classes: Array<string>) => {
-  classes.forEach(cls => element.classList.toggle(cls));
+export const toggleMultipleClasses = (
+  element: HTMLElement,
+  ...classes: Array<string>
+) => {
+  classes.forEach((cls) => element.classList.toggle(cls));
```
lxsmnsyc, line 119: `for..of` is fine. Honestly let's just keep from using `forEach` on all iterables

RobPruzan: just formatting isn't my code

packages/scan/src/core/web/utils/outline.ts
```diff
@@ -391,178 +301,281 @@ export const fadeOutOutline = (
...
-) => {
-  const newPreviousOutlines = new Map<string, PendingOutline>();
+const activateOutlines = async () => {
+  const domNodes: Array<HTMLElement> = [];
+  const scheduledOutlines = ReactScanInternals.scheduledOutlines;
+  const activeOutlines = ReactScanInternals.activeOutlines;
+  const activeFibers = new Map<Fiber, AggregatedRender>();
+
+  for (const activeOutline of ReactScanInternals.activeOutlines.values()) {
+    activeOutline.groupedAggregatedRender?.forEach(
```
lxsmnsyc, line 351: `for..of` here

packages/scan/src/core/web/utils/helpers.ts
RobPruzan: same thing as other comment, just formatting not my code

packages/scan/src/core/web/utils/helpers.ts
lxsmnsyc: oof then

PR #157, opened 2024-12-20
fix: Monitoring Correctness

NisargIO:
1. performance
2. data correctness
3. still waiting on interaction metadata to fix interactions being negative / too high

### DIFF
--- a/packages/scan/src/core/monitor/index.ts
+++ b/packages/scan/src/core/monitor/index.ts
@@ -1,5 +1,5 @@
 'use client';
-import { getDisplayName, isCompositeFiber } from 'bippy';
+import { getDisplayName, getTimings, isCompositeFiber } from 'bippy';
 import { type Fiber } from 'react-reconciler';
 import { useEffect } from 'react';
 import {
@@ -73,9 +73,7 @@ export const Monitoring = ({
   }
 
   useEffect(() => {
-    scanMonitoring({
-      enabled: true,
-    });
+    scanMonitoring({ enabled: true });
     return initPerformanceMonitoring();
   }, []);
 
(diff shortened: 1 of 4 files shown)


packages/scan/src/core/monitor/index.ts
```diff
@@ -149,44 +146,41 @@ const aggregateComponentRenderToInteraction = (
...
   const monitor = Store.monitor.value;
-  if (monitor && monitor.interactions && monitor.interactions.length > 0) {
-    const latestInteraction =
-      monitor.interactions[monitor.interactions.length - 1];
-
-    let totalTime = 0;
-    for (const render of renders) {
-      totalTime += render.time ?? 0;
-    }
+  if (!monitor || !monitor.interactions || monitor.interactions.length === 0)
```
RobPruzan, line 149: don't have to make this change, but for weird stuff like this in the future just comment 

`// invariant: monitor must exist when calling this function`

one of the few type of comments i find useful and makes future cleanups easier

NisargIO: That conditional was there before, just made it a early return

packages/scan/src/core/monitor/index.ts
```diff
@@ -149,44 +146,41 @@ const aggregateComponentRenderToInteraction = (
...
+  component.renders += rendersCount;
 
-    component.renders += renders.length;
-    if (!component.totalTime) {
-      component.totalTime = 0;
-    }
-    component.totalTime += component.totalTime
-      ? component.totalTime + totalTime
-      : totalTime;
+  // We leave the times undefined to differentiate between a 0ms render and a non-profiled render.
```
RobPruzan, line 178: doesn't the json replacer kill properties with undefined values?

RobPruzan (approved):
lgtm, merge when ready and ill test everything e2e before a release

packages/scan/src/core/monitor/index.ts
RobPruzan: yeah not trying to nit, just thought was a good opportunity to say that

packages/scan/src/core/monitor/index.ts
NisargIO: Yes but in clock house it can be nullable

PR #164, opened 2024-12-25
Update README.md, use next script tag

21e8:
Hi,

I had some problems loading with the instructions given in the readme.
This seems to work better for me on nextjs.

Regards

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,8 @@ export default function Document() {
 Add the script tag to your `app/layout.tsx`:
 
 ```jsx
+import Script from "next/script";
+
 export default function RootLayout({
   children,
 }: {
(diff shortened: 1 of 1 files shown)


heyglassy:
> Hi,
> 
> I had some problems loading with the instructions given in the readme. This seems to work better for me on nextjs.
> 
> Regards

For some reason Arc wasn't displaying the re-renders but was loading the script. This ended up fixing that issue. Thank you!

kewldan:
If you don't specify strategy at all, the script will run normally, thus not increasing the “initial HTML from the server” following [Next.JS Docs](https://nextjs.org/docs/app/api-reference/components/script). Also you don't have to specify Script in the head tag, because in your case, when beforeInteractive is specified, Script will be automatically moved to the head, also following the above documentation

PR #166, opened 2024-12-25
refactor: props, states and contexts view

pivanov:
https://github.com/user-attachments/assets/d1a76077-5175-4ec3-8430-8056e6d6b480


packages/scan/src/core/index.ts
```diff
@@ -533,23 +536,30 @@ export const start = () => {
...
+      // Create audio context on first user interaction
+      const createAudioContextOnInteraction = () => {
+        audioContext = new (
+          window.AudioContext ||
+          // @ts-expect-error -- This is a fallback for Safari
+          window.webkitAudioContext
+        )();
+
+        void audioContext.resume();
+        window.removeEventListener('pointerdown', createAudioContextOnInteraction);
```
lxsmnsyc, line 557: you probably don't need this since you set the listener to `once`

packages/scan/src/core/instrumentation.ts
```diff
@@ -205,72 +202,49 @@ export const getPropsChanges = (fiber: Fiber) => {
...
+    };
+    changes.push(change);
 
-export const getContextChanges = (fiber: Fiber) => {
-  const changes: Array<RenderChange> = [];
+    const prevValueString = fastSerialize(prevValue);
+    const nextValueString = fastSerialize(nextValue);
 
-  traverseContexts(fiber, getContextChangesTraversal.bind(changes));
+    if (
```
lxsmnsyc, line 240: oof, I can move this back to bound functions.

packages/scan/src/core/instrumentation.ts
```diff
@@ -311,56 +285,43 @@ let inited = false;
...
   for (const mutatedHostFiber of mutatedHostFibers) {
-    const state: IsRenderUnnecessaryState = {
-      isRequiredChange: false,
-    };
-    traverseProps(mutatedHostFiber, isRenderUnnecessaryTraversal.bind(state));
-    if (state.isRequiredChange) return false;
+    let isRequiredChange = false;
+    traverseProps(mutatedHostFiber, (prevValue, nextValue) => {
+      if (
+        !Object.is(prevValue, nextValue) &&
```
lxsmnsyc, line 297: please use `isEqual`

packages/scan/src/core/index.ts
pivanov: oh ... my bad you are absolutely right

packages/scan/src/core/instrumentation.ts
pivanov: Oh, correct again! Thanks so much ... I'll use this in a few more places to stay consistent.

packages/scan/src/core/web/inspect-element/view-state/utils.ts
```diff
@@ -0,0 +1,380 @@
...
+
+  return contexts;
+};
+
+export const getCurrentContext = (fiber: Fiber) => {
+  const contexts = getAllFiberContexts(fiber);
+  const contextObj: Record<string, unknown> = {};
+
+  contexts.forEach((value, contextName) => {
+    contextObj[contextName] = value.displayValue;
```
lxsmnsyc, line 336: Take note: in the future we might want to change this implementation, as this is a megamorphic code (hurts perf)

packages/scan/src/core/web/inspect-element/view-state/index.ts
```diff
@@ -0,0 +1,852 @@
...
+let changedAtInterval: ReturnType<typeof setInterval> | null = null;
+
+export const changedAt = new Map<string, number>();
+const lastRendered = new Map<string, unknown>();
+
+const templates = {
+  whatChangedSection: createHTMLTemplate<HTMLDetailsElement>(
+    `<details class="react-scan-what-changed" style="background-color:#b8860b;color:#ffff00;padding:5px">
+      <summary class="font-bold">What changed?</summary>
+    </details>`,
```
lxsmnsyc, line 50: quotes and closing tags are optional, but this is fine.

packages/scan/src/core/web/inspect-element/view-state/index.ts
```diff
@@ -0,0 +1,852 @@
...
+          changedProps,
+        ),
+        hasChanges: changedProps.size > 0,
+      });
+    }, null);
+  }
+
+  const currentContext = getCurrentContext(fiber);
+  if (Object.keys(currentContext).length) {
+    tryOrElse(() => {
```
lxsmnsyc, line 243: in the future, it should be of preference that we should use `try-catch` directly instead of using utilities like this (for 2 reasons: JS compilers cannot optimize this + closures are painful to optimize with). The utility itself has some ideal uses, but not in this case, I guess. Don't worry, I'll fix this one.

packages/scan/src/core/web/inspect-element/view-state/index.ts
pivanov: Oh yeah ... After I understood all the logic and sorted it out a bit, I think the next step is to migrate to preact.

packages/scan/src/core/web/inspect-element/view-state/utils.ts
pivanov: Note taken!

packages/scan/src/core/web/inspect-element/view-state/index.ts
lxsmnsyc: yeah we could move to it entirely. Your preference though!

packages/scan/src/core/web/inspect-element/view-state/utils.ts
```diff
@@ -0,0 +1,380 @@
...
+  if (!fiber.alternate) return changes;
+
+  const currentContexts = getAllFiberContexts(fiber);
+
+  currentContexts.forEach((_currentValue, contextType) => {
+    const contextName = (typeof contextType === 'object' && contextType !== null)
+      ? (contextType as any)?.displayName ??
+      (contextType as any)?.Provider?.displayName ??
+      (contextType as any)?.Consumer?.displayName ??
+      (contextType as any)?.type?.name?.replace('Provider', '') ??
```
lxsmnsyc, line 353: I just noticed there's so many things going on here (mainly just too many `?`), perhaps we can move this to a separate module level function? just for it to be self-descriptive.

pivanov: Nice one! Now we have `getContextDisplayName` :)

PR #168, opened 2024-12-25
fix: dependencies

lxsmnsyc:
This fixes some weird dependencies `react-scan` has that shouldn't be there in the first place


packages/scan/package.json
```diff
@@ -233,8 +229,11 @@
     "@types/react": "^18.0.0",
     "@types/react-reconciler": "^0.28.8",
     "@types/react-router": "^5.1.0",
+    "@types/node": "^20.17.9",
```
pivanov, line 232: Can we add this with `pnpm add -Dw @types/node` we need it in all workspaces.

lxsmnsyc: well it doesn't have to be on all packages.

PR #177, opened 2025-01-02
feat: preact props, states and contexts

pivanov:


packages/scan/src/web/components/inspector/index.tsx
```diff
@@ -0,0 +1,1358 @@
...
+  return Object.keys(value).length > 0;
+};
+
+const isPromise = (value: any): value is Promise<unknown> => {
+  return value && (value instanceof Promise || (typeof value === 'object' && 'then' in value));
+};
+
+const isEditableValue = (value: unknown): boolean => {
+  // Basic types are always editable
+  if (typeof value === 'string' ||
```
lxsmnsyc, line 112: hmmm, I think `switch` is a better alternative here.

packages/scan/src/web/components/inspector/index.tsx
```diff
@@ -0,0 +1,1358 @@
...
+    typeof value === 'number' ||
+    typeof value === 'boolean' ||
+    typeof value === 'bigint') return true;
+
+  if (value === null || value === undefined) return true;
```
lxsmnsyc, line 117: `if (value == null)` is the shorter alternative. You could also put this before the `typeof` check as a better short-circuit.

packages/scan/src/web/components/inspector/index.tsx
```diff
@@ -0,0 +1,1358 @@
...
+
+  // Special objects
+  if (value instanceof Date ||
+    value instanceof RegExp ||
+    value instanceof Error) return true;
```
lxsmnsyc, line 122: Note: `.constructor` is faster here, but I'll let you leave this as-is. Let's change this in another time.

packages/scan/src/web/components/inspector/index.tsx
```diff
@@ -0,0 +1,1358 @@
...
+  const currentPath = getPath(
+    getDisplayName(fiber.type) ?? 'Unknown',
+    section,
+    parentPath ?? '',
+    name
+  );
+  const prevValue = lastRendered.get(currentPath);
+  const isChanged = prevValue !== undefined && !isEqual(prevValue, value);
+
+  const renderNestedProperties = useCallback((obj: Record<string, unknown> | Array<unknown> | Map<unknown, unknown> | Set<unknown> | ArrayBuffer | DataView | Int8Array | Uint8Array | Uint8ClampedArray | Int16Array | Uint16Array | Int32Array | Uint32Array | Float32Array | Float64Array | BigInt64Array | BigUint64Array) => {
```
lxsmnsyc, line 706: oof. can we do a `type` here.

packages/scan/src/web/components/inspector/index.tsx
```diff
@@ -0,0 +1,1358 @@
...
+          }
+        }
+      } else if (obj instanceof Set) {
+        for (const value of obj) {
+          if (hasCircular(value)) {
+            return true;
+          }
+        }
+      } else if (Array.isArray(obj)) {
+        for (const value of obj) {
```
lxsmnsyc, line 833: can probably simplify this and `Set` check into a single if block

packages/scan/src/web/components/inspector/index.tsx
pivanov: oh ... it's because of the eslint `eqeqeq`

packages/scan/src/web/components/inspector/overlay/utils.ts
```diff
@@ -74,116 +69,87 @@ export const getStateNames = (fiber: Fiber): Array<string> => {
...
 
+// Helper to check if a component is a direct component (not a wrapper/HOC)
+export const isDirectComponent = (fiber: Fiber): boolean => {
+  if (!fiber || !fiber.type) return false;
+
+  // Check if it's a functional component or a class component
+  const isFunctionalComponent = typeof fiber.type === 'function';
+  const isClassComponent = fiber.type.prototype && fiber.type.prototype.isReactComponent;
+
+  if (!isFunctionalComponent && !isClassComponent) return false;
```
lxsmnsyc, line 80: `!(isFunctionalComponent || isClassComponent)`

fun fact: biome actually has a linter feature that simplifies boolean logic, but we can't use it right now 🤣

packages/scan/src/web/components/inspector/utils.ts
```diff
@@ -224,3 +282,86 @@ export const getOverrideMethods = (): OverrideMethods => {
 
   return { overrideProps, overrideHookState };
 };
+
+
+export const nonVisualTags = [
```
lxsmnsyc, line 287: we probably want this as a `Set`

packages/scan/src/web/components/widget/index.tsx
```diff
@@ -334,81 +316,85 @@ export const Widget = () => {
...
-            'transition-colors duration-200',
-            'overflow-hidden',
-            'rounded-lg',
-            'z-10'
-          )}
-        >
-          <Toolbar refPropContainer={refPropContainer} />
+          <div
+            ref={refFooter}
+            className={cn(
```
lxsmnsyc, line 384: hmmm JSX has a feature that allows string properties to support new lines. I mean this is fine as well, for preferences.

packages/scan/src/web/components/inspector/overlay/utils.ts
pivanov: huh yeah :) 
De Morgan's Law ...

packages/scan/src/web/components/widget/index.tsx
pivanov: yeah ... but the output is not pretty :) 
```
... children:Xn(eo,{})})]}),Xn("div",{ref:r,className:"\n              relative\n              z-10\n              flex h-9 items-center\n              justify-between overflow-hidden\n              rounded-lg\n              transition-colors\n              duration-200\n            ",ch ...
```
that's why I keep it like this ... to save few bytes

packages/scan/src/web/components/inspector/index.tsx
lxsmnsyc: ah I forgot that exists. I guess we need to tweak it: https://eslint.org/docs/latest/rules/eqeqeq#always

pivanov: yeah I'm just not sure about the tweak :) how to be not so restrictive and annoying at the same time 🤣

packages/scan/src/web/components/inspector/index.tsx
```diff
@@ -108,21 +123,25 @@ const isPromise = (value: any): value is Promise<unknown> => {
...
-    typeof value === 'bigint') return true;
-
   if (value === null || value === undefined) return true;
 
-  // Special objects
-  if (value instanceof Date ||
-    value instanceof RegExp ||
-    value instanceof Error) return true;
+  switch (true) {
+    case value?.constructor === Date:
```
lxsmnsyc, line 129: this shouldn't be nullish anymore right? since there's a typeguard before this one. Also, `switch (value.constructor)` might be simpler.

pivanov: oh ... absolutely ... eagle eye!!!

PR #186, opened 2025-01-09
refactor: migrate from eslint to biome and improve architecture

pivanov:
## TODOS
@RobPruzan 
- [ ] port new outline logic over
- [ ] implement settings

@pivanov
- [x] fix: the purple animated flash on state change, it spazzes out when things change frequently, that shouldn’t happen
- [x] fix: bug causing state updates to show every other change
- [x] fix: whats changed shouldn’t show if nothing changed
- [x] fix: overlay cursor should be always the default
- [x] fix: handle missing fiber when inspecting elements
- [x] fix: prevent inspecting the toolbar widget
- [x] fix: toolbar widget animations
- [x] fix: what changed metrics
- [x] fix: replay button behaviour 
- [x] feat: remove audio stuff entirely in favore of settings
- [x] feat: add provider names to context area
- [x] feat: inspect timeline (@TODO: @pivanov improve UI and finish the implementation)
- [x] cleanup old CSS stuff
- [x] Improve UX/UI/DX: Display props with the following visual cues:
  - 🛡️ Memoized props (safe and optimized).
  - 🔥 Non-memoized function props (potential performance issue).
  - 🔔 Props requiring observation (watch closely).
- [x] improve: overlay fade in/out instead of flick
- [x] improve: use bippy `bippy 0.2.0`
- [x] improve: remove `react-reconciler` package in favor of `bippy`
- [x] chore: replace ESLint with Biome for linting and formatting
- [x] chore: configure Biome with strict rules for:
  - Type checking and type safety
  - Code correctness (unused imports, variables, parameters)
  - Security (noExplicitAny, noConsole, dangerouslySetInnerHTML)
- [x] improve VSCode integration:
  - Set up Biome formatters for TS, JS, JSON, MD, CSS
  - Add automatic code actions and formatting
- [x] Improve React internals:
  - Remove react-reconciler dependency
  - Use bippy for Fiber/FiberRoot types
  - Use installRDTHook for React DevTools hook to avoid side effects

@RobPruzan & @pivanov 
- [ ] audit everything, make sure its 100% stable


packages/scan/src/core/monitor/network.ts
```diff
@@ -35,16 +35,17 @@ export const flush = async (): Promise<void> => {
     const timeSinceStart = now - interaction.performanceEntry.startTime;
     // these interactions were retried enough and should be discarded to avoid mem leak
     if (timeSinceStart > 30000) {
-      continue;
+      // Skip this iteration
```
NisargIO, line 38: Why was the continue removed?

packages/scan/src/core/monitor/performance.ts
```diff
@@ -151,7 +155,7 @@ const getFirstNamedAncestorCompositeFiber = (element: Element) => {
     if (!fiber) {
       continue;
     }
-    if (fiber.type && getDisplayName(fiber.type)) {
+    if (getDisplayName(fiber?.type)) {
```
NisargIO, line 158: ```fiber.type &&``` should not be removed here

packages/scan/src/web/utils/outline-worker.ts
```diff
@@ -45,7 +45,7 @@ function setupOutlineWorker(): (action: OutlineWorkerAction) => Promise<void> {
     'Menlo,Consolas,Monaco,Liberation Mono,Lucida Console,monospace';
   let ctx: OffscreenCanvasRenderingContext2D | undefined;
 
-  const enum Reason {
+  enum Reason {
```
lxsmnsyc, line 48: hmmm, is this a linter change? it was specifically meant to be `const enum` so that the values are inlined. TypeScript's `enum` output is bloated.

packages/scan/src/core/monitor/network.ts
pivanov: autosafe biome .... :(
sorry
fixed to:
```
for (let i = 0; i < interactions.length; i++) {
    const interaction = interactions[i];
    const timeSinceStart = now - interaction.performanceEntry.startTime;
    if (timeSinceStart <= 30000) {
      // Skip interactions older than 30 seconds to prevent memory leaks
      if (timeSinceStart <= INTERACTION_TIME_TILL_COMPLETED) {
        pendingInteractions.push(interaction);
      } else {
        completedInteractions.push(interaction);
      }
    }
  }
  ```

File: packages/scan/src/core/monitor/performance.ts (hunk 2 above)
Comment (pivanov) [reply packages/scan/src/core/monitor/performance.ts:158]: fixed

File: packages/scan/src/web/utils/outline-worker.ts (hunk 3 above)
Comment (pivanov) [reply packages/scan/src/web/utils/outline-worker.ts:48]: yeah ... biome autosafe issue :( .. fixed

File: packages/scan/src/core/monitor/performance.ts (hunk 2 above)
Comment (lxsmnsyc) [reply packages/scan/src/core/monitor/performance.ts]: btw I made this change. Previously it was the latter but I made it so that it doesn't have to call `getDisplayName` unnecessarily.

File: packages/scan/src/web/utils/outline-worker.ts (hunk 3 above)
Comment (lxsmnsyc) [reply packages/scan/src/web/utils/outline-worker.ts]: ouch.

Comment (pivanov) [reply packages/scan/src/web/utils/outline-worker.ts]: yeah ... we will need more 👀

File: .vscode/settings.json [hunk 4]
```diff
@@ -1,20 +1,24 @@
...
+    "editor.defaultFormatter": "biomejs.biome"
+  },
+  "[javascript]": {
+    "editor.defaultFormatter": "biomejs.biome"
+  },
+  "[json]": {
+    "editor.defaultFormatter": "biomejs.biome"
+  },
+  "[markdown]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
```
lxsmnsyc, line 18: hmmm honestly we probably don't want to set this down (or maybe we could, markdownlint is a thing after all)

biome.json
```diff
@@ -7,28 +7,69 @@
...
+    "indentWidth": 2,
+    "lineWidth": 80
   },
   "javascript": {
     "formatter": {
       "quoteStyle": "single",
-      "indentStyle": "space",
-      "indentWidth": 2,
-      "lineWidth": 80,
-      "lineEnding": "lf"
```
lxsmnsyc, line 31: we need this because Windows sucks

packages/scan/src/core/index.ts
```diff
@@ -752,15 +784,13 @@ export const onRender = (
...
-    React.ReactNode,
-    undefined | null | string | number | boolean | bigint
-  >
+  Exclude<ReactNode, undefined | null | string | number | boolean | bigint>
 >();
 
-export const ignoreScan = (node: React.ReactNode) => {
-  if (typeof node === 'object' && node) {
+export const ignoreScan = (node: ReactNode) => {
+  if (typeof node === 'object' && node !== null) {
```
lxsmnsyc, line 793: no need for null checks. Optional change.

packages/scan/src/core/instrumentation.ts
```diff
@@ -53,7 +54,7 @@ export const isElementVisible = (el: Element) => {
   return (
     style.display !== 'none' &&
     style.visibility !== 'hidden' &&
-    (style as any).contentVisibility !== 'hidden' &&
+    style.getPropertyValue('content-visibility') !== 'hidden' &&
```
lxsmnsyc, line 57: weird.

packages/scan/src/core/instrumentation.ts
```diff
@@ -81,7 +82,7 @@ export const isElementInViewport = (
   return isVisible && rect.width && rect.height;
 };
 
-export const enum ChangeReason {
+export enum ChangeReason {
```
lxsmnsyc, line 85: needs to be `const enum`

packages/scan/src/web/components/inspector/overlay/utils.ts
```diff
@@ -38,21 +36,32 @@ const ensureRecord = (
...
+
+    case Boolean(
+      value &&
+        (value instanceof Promise ||
+          (typeof value === 'object' && 'then' in value)),
+    ):
+      return { type: 'promise' };
+
+    case typeof value === 'object': {
+      if (seen.has(value as object)) {
```
lxsmnsyc, line 64: Previous implementation is intentional. The reason is so that we can short-circuit via `typeof value === 'object'`. `instanceof` is slow and switch-case is a fallthrough so there are some unnecessary flow of code here.

packages/scan/src/web/components/inspector/overlay/utils.ts
```diff
@@ -62,14 +71,14 @@ const ensureRecord = (
         return { type: 'array', length: value.length, items: safeArray };
       }
 
-      seen.add(value);
+      seen.add(value as object);
```
lxsmnsyc, line 74: ... and we don't have to do this

packages/scan/src/web/components/inspector/utils.ts
```diff
@@ -300,39 +327,36 @@ export const getOverrideMethods = (): OverrideMethods => {
...
-  'OBJECT',
-  'PARAM',
-  'SOURCE',
-  'TRACK',
-  'AREA',
-  'PORTAL',
-  'SLOT',
-  'XML',
-  'DOCTYPE',
-  'COMMENT'
```
lxsmnsyc, line 324: nothing was wrong here btw

.vscode/settings.json
pivanov: Yeah I think is better to have something than nothing and we can change it at any time

packages/scan/src/core/index.ts
```diff
@@ -693,16 +725,16 @@ export const start = () => {
...
       console.error(
         '[React Scan] Failed to load. Must import React Scan before React runs.',
       );
     }, 5000);
   }
 };
 
-export const withScan = <T>(
-  component: React.ComponentType<T>,
+export const withScan = <T extends {}>(
```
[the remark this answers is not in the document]
pivanov: autofix makes me crazy some time

biome.json
```diff
@@ -7,28 +7,69 @@
...
-    "indentWidth": 2,
-    "lineWidth": 80,
-    "lineEnding": "lf"
+    "ignore": [
+      "examples/**",
+      "**/dist/**",
+      "**/build/**",
+      "node_modules",
+      "**/node_modules/**",
+      "**/*.css",
```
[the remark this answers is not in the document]
pivanov: About the CSS ... unfortunately ... It struggles with modern CSS features like @-moz-document
About the `examples` ... I will fix everything there in another PR because I think we should move the website from `packages` to `examples` too

packages/scan/src/core/instrumentation.ts
pivanov: fix this before biome ... was an issue with eslint ... I will fix it with `style.contentVisibility`

packages/scan/src/web/components/inspector/utils.ts
pivanov: ahhh stupid me! copy paste mistake :(

RobPruzan:
TODOS:
Rob:
  - port new outline logic over
  - implement settings
  - audit
 Pavel (excluding what's mentioned in PR)
   - remove audio stuff entirely
   - if you have time and it's easy look into the animation of the toolbar (make sure it always animates smoothly, and make the animations a little smoother- this is probably as easy as cursoring it for 20 mins


This __should__ be, including your list, the exhaustive set of things we need to get done

packages/scan/src/core/index.ts
lxsmnsyc: no I mean, you can just do a `&& node` no?

biome.json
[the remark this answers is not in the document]
lxsmnsyc: > About the CSS ... unfortunately ... It struggles with modern CSS features like @-moz-document

oh oof, I guess we can ignore then.

packages/scan/src/core/index.ts
pivanov: oh yeah ... we can do this

lxsmnsyc (approved):
LGTM

pivanov:
obsolate

PR #189, opened 2025-01-15
fix: react component names + turbopack loader

NisargIO:


packages/scan/package.json
```diff
@@ -153,6 +153,11 @@
       "types": "./dist/react-component-name/astro.d.ts",
       "import": "./dist/react-component-name/astro.mjs",
       "require": "./dist/react-component-name/astro.js"
+    },
+    "./react-component-name/loader": {
```
RobPruzan, line 157: add "typesVersions" for this, reference the others, its to support older node versions

Issue #178, opened 2025-01-02
Make the client library separate for much smaller install 


natew:
One.js is now shipping React scan, but it massively increases or npm install size. And since we use it for both our runtime and as our CLI install (npx one) it slows down getting started a lot too. 

This is because there’s a bunch of backend type dops included like playwright. 

Creating a new “@react-scan/client” library would greatly help this. Would love to see this, or we may need to split it out into a separate plugin until then. I can maybe help a bit with the initial split if you are strapped for time.

RobPruzan:
This is a priority for us in the next release

PR #193, opened 2025-01-17
chore(scan): remove unused dependencies

43081j:
Removes unused dependencies from the `scan` package:

- `@babel/generator` - not used directly, already pulled in via `@babel/core`
- `@clack/core` - not used directly, already pulled in via `@clack/prompts`
- `estree-walker` - not used
- `tsx` - not used

At the root of the monorepo:

- `@vercel/style-guide` - doesn't seem to be referenced anywhere

Also changes a couple of FS paths to use `__dirname` in case `tsup` is ever run from another `cwd`.

This should drop a big chunk of the install size, re #178. Specifically since we drop `tsx` which is ~400KB.


RobPruzan (approved):
looks good, going to rebase and resolve conflicts for you so we can get this merged

43081j:
@NisargIO i noticed this got reverted. any idea why? doubtful we need all those unused deps to fix whatever problem came up

happy to fix it properly if you can point me at the issue

RobPruzan:
going to un-revert the revert, was an issue with catching up the branch when merging

PR #200, opened 2025-01-22
Replays for React Scan

RobPruzan:
This PR Introduces a runtime for react scan to:
- collect detailed interaction timing by hacking details from browser event loop callbacks
- associate the custom interaction timing tracking with the PerformanceEntry API to derive the time between commit end and new frame presentation
- collect rrweb session replay events
- runs custom rrweb plugins that:
  - collect react-scan data, allowing the render outlines to be replayed without the user seeing them
  - collect live fps updates (for the purpose of correlating them to the replay)
  - track interaction timings during the session replay

Other changes include:
- an rrweb replayer plugin to render a canvas over the session replay with react scan outlines
- updated render tracking used in react scan monitoring
  - we now only track the blocking component renders
- collect the fiber subtree that re-rendered when ingesting in monitoring
- also collect the subtree of interacted element, incase the target was not the leaf node

What the PR does not include is the infrastructure behind replays for react scan, that will be open sourced at a future date

That functionaliy roughly includes:
- an ingestion server to collect interaction data
- a puppeteer cluster to replay the session replay offline, this allows us to:
  -  convert the session replay to an mp4 by taking successive screenshots of the replayed session
  - take screenshots of the element interacted with
  - take screenshots of the render outlines over the page when the interaction happened
  - run expensive dom queries to understand the state of the app that we can't run on the users device
    - for example, querying for every single dom element if its in the viewport. This is not feasible to run on user device, but is trival to run offline in puppeteer
- storage for replays as mp4's, element screenshots, and video thumbnails
- Interaction summary w/ LLM's **without** using personalized user data
  - we never send data about the state of the application to a third party. This means we are limited in how good the summaries can get using only "static" information about the fiber tree. But, with a good enough prompt, the model does a good job in the general case

## Preview of visualized collected data

https://github.com/user-attachments/assets/a6d39b27-991e-4992-8d64-e9952002ab3d


packages/scan/package.json
```diff
@@ -1,8 +1,14 @@
 {
   "name": "react-scan",
-  "version": "0.0.54",
+  "version": "0.0.1083",
```
RobPruzan, line 3: this was incremented every time i tested a new change...

packages/scan/src/core/monitor/performance.ts
```diff
@@ -321,3 +496,720 @@ const setupPerformanceListener = (
...
+    | TimeoutStage
+    | UnInitializedStage
+  ) & { stageStart: number };
+};
+
+/**
+ *
+ * handles tracking event timings for arbitrarily overlapping handlers with cancel logic
+ */
+export const setupDetailedPointerTimingListener = (
```
RobPruzan, line 691: where the magic happens

packages/scan/src/core/index.ts
```diff
@@ -461,7 +466,12 @@ const updateScheduledOutlines = (fiber: Fiber, renders: Array<Render>) => {
   for (let i = 0, len = renders.length; i < len; i++) {
     const render = renders[i];
     const domFiber = getNearestHostFiber(fiber);
-    if (!domFiber || !domFiber.stateNode || !(domFiber.stateNode instanceof Element)) continue;
+    if (
+      !domFiber ||
+      !domFiber.stateNode ||
+      !(domFiber.stateNode instanceof Element)
+    )
```
lxsmnsyc, line 473: De Morgan's Law 😉

packages/scan/src/core/monitor/session-replay/record.ts
```diff
@@ -0,0 +1,1182 @@
...
+const initRRWeb = async () => {
+  if (!record) {
+    const rrwebPkg = await import('rrweb');
+    record = rrwebPkg.record;
+  }
+  return record;
+};
+
+export const startNewRecording = async () => {
+  console.log('new recording start');
```
lxsmnsyc, line 108: removable?

packages/scan/src/core/monitor/session-replay/record.ts
```diff
@@ -0,0 +1,1182 @@
...
+  /* This value is computed before the full rendered text is shown, so its only considered an estimate */
+  estimatedTextWidth?: number; // todo: estimated is stupid just make it the actual
+
+  alpha?: number | null;
+  totalFrames?: number | null;
+}>;
+
+let lastFiberId = 0;
+const fiberIdMap = new Map<Fiber, number>();
+const getFiberId = (fiber: Fiber) => {
```
lxsmnsyc, line 158: how different is this from bippy's getFiberID

packages/scan/src/core/monitor/session-replay/record.ts
```diff
@@ -0,0 +1,1182 @@
...
+  completedAt: number;
+};
+
+export const flushInvariant = (payload: SerializableFlushPayload) => {
+  // devInvariant(
+  //   payload.events[0].type === 4,
+  //   "first rrweb event must be meta event"
+  // );
+  if (payload.events[0].type !== 4) {
+    console.warn('hm', payload);
```
lxsmnsyc, line 274: we probably need a DEV flag here now lol, perhaps `import.meta.env.DEV`

packages/scan/src/core/monitor/session-replay/record.ts
RobPruzan: this PR is not in a reviewable state yet (there are probably on the order of 1000 comments that could be left)

only opening it for visibility

packages/scan/src/core/monitor/session-replay/record.ts
RobPruzan: i think i wrote that before it existed

packages/scan/src/index.ts
```diff
@@ -1,3 +1,5 @@
 import 'bippy'; // implicit init RDT hook
 
 export * from './core/index';
+
+export { ReactScanReplayPlugin } from './core//monitor/session-replay/replay-v2';
```
imcodingideas, line 5: Should we remove the extra `/` here?

packages/scan/src/core/monitor/performance-store.ts
```diff
@@ -0,0 +1,115 @@
...
+  publish: (
+    item: Item,
+    to: ChanelName,
+    dropFirst: boolean,
+    createIfNoChannel: boolean,
+  ) => void;
+  channels: Record<
+    ChanelName,
+    { callbacks: BoundedArray<Callback>; state: BoundedArray<Item> }
+  >;
```
imcodingideas, line 25: I think you can do something like this, and use generics here.

```suggestion
  channels: Record<ChannelName, Channel<T>>;
```

and you could create an interface for it like this.. 

```
interface Channel<T> {
  callbacks: BoundedArray<Callback<T>>;
  state: BoundedArray<T>;
}
```

packages/scan/src/core/monitor/performance-store.ts
```diff
@@ -0,0 +1,115 @@
...
+  getAvailableChannels: () => BoundedArray<string>;
+  updateChannelState: (
+    channel: ChanelName,
+    updater: Updater,
+    createIfNoChannel: boolean,
+  ) => void;
+};
```
imcodingideas, line 32: This is a little more strongly typed here.. and you can implement `PerformanceEntryChannelsType` 

```suggestion
export type ChannelItem = InternalInteraction | PerformanceInteraction;
export type UnSubscribe = () => void;
export type Callback<T> = (item: T) => void;
export type Updater<T> = (state: BoundedArray<T>) => BoundedArray<T>;
export type ChannelName = string;

interface Channel<T> {
  callbacks: BoundedArray<Callback<T>>;
  state: BoundedArray<T>;
}

interface PerformanceEntryChannelsType<T> {
  subscribe: (to: ChannelName, cb: Callback<T>, dropFirst?: boolean) => UnSubscribe;
  publish: (
    item: T,
    to: ChannelName,
    createIfNoChannel?: boolean
  ) => void;
  channels: Record<ChannelName, Channel<T>>;
  getAvailableChannels: () => BoundedArray<string>;
  updateChannelState: (
    channel: ChannelName,
    updater: Updater<T>,
    createIfNoChannel?: boolean
  ) => void;
  getChannelState: (channel: ChannelName) => BoundedArray<T>;
}
```

packages/scan/src/core/monitor/performance-store.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    existingChannel.state = updater(existingChannel.state);
+  }
+
+  getChannelState(channel: ChanelName) {
+    return this.channels[channel].state ?? new BoundedArray<Item>(MAX_CHANNEL_SIZE);
+  }
+}
+
+export const performanceEntryChannels = new PerformanceEntryChannels();
```
imcodingideas, line 115: This is really great. I'm so excited for this feature. 

Fly on the wall here without write access. Happy to help with TypeScript a little I don't step on your toes.


```suggestion
class PerformanceEntryChannels<T> implements PerformanceEntryChannelsType<T> {
  private readonly channels: Record<ChannelName, Channel<T>> = {};

  private createChannel(): Channel<T> {
    return {
      callbacks: new BoundedArray<Callback<T>>(MAX_CHANNEL_SIZE),
      state: new BoundedArray<T>(MAX_CHANNEL_SIZE),
    };
  }

  private ensureChannel(channelName: ChannelName, createIfNoChannel = true): Channel<T> | null {
    if (!this.channels[channelName]) {
      if (!createIfNoChannel) {
        return null;
      }
      this.channels[channelName] = this.createChannel();
    }
    return this.channels[channelName];
  }

  public publish(item: T, to: ChannelName, createIfNoChannel = true): void {
    const channel = this.ensureChannel(to, createIfNoChannel);
    if (!channel) return;

    channel.state.push(item);
    channel.callbacks.forEach((cb) => {
      try {
        cb(item);
      } catch (error) {
        console.error(`Error in channel "${to}" callback:`, error);
      }
    });
  }

  public subscribe(to: ChannelName, cb: Callback<T>, dropFirst = false): UnSubscribe {
    const channel = this.ensureChannel(to, true);
    if (!channel) throw new Error(`Failed to create/access channel: ${to}`);

    channel.callbacks.push(cb);

    if (!dropFirst) {
      channel.state.forEach((item) => {
        try {
          cb(item);
        } catch (error) {
          console.error(`Error in channel "${to}" subscription callback:`, error);
        }
      });
    }

    return () => {
      const filtered = channel.callbacks.filter(
        (subscribed) => subscribed !== cb
      );
      channel.callbacks = BoundedArray.fromArray(filtered, MAX_CHANNEL_SIZE);
    };
  }

  public getAvailableChannels(): BoundedArray<string> {
    return BoundedArray.fromArray(Object.keys(this.channels), MAX_CHANNEL_SIZE);
  }

  public updateChannelState(
    channelName: ChannelName,
    updater: Updater<T>,
    createIfNoChannel = true
  ): void {
    const channel = this.ensureChannel(channelName, createIfNoChannel);
    if (!channel) return;

    try {
      channel.state = updater(channel.state);
    } catch (error) {
      console.error(`Error updating channel "${channelName}" state:`, error);
    }
  }

  public getChannelState(channelName: ChannelName): BoundedArray<T> {
    const channel = this.channels[channelName];
    return channel?.state ?? new BoundedArray<T>(MAX_CHANNEL_SIZE);
  }

  public clearChannel(channelName: ChannelName): void {
    const channel = this.channels[channelName];
    if (channel) {
      channel.state = new BoundedArray<T>(MAX_CHANNEL_SIZE);
    }
  }

  public removeChannel(channelName: ChannelName): void {
    delete this.channels[channelName];
  }

  public getSubscriberCount(channelName: ChannelName): number {
    return this.channels[channelName]?.callbacks.length ?? 0;
  }
}

RobPruzan: thanks for the contributions, for context I will probably be migrating this logic over for a react-scan feature (some harder backend problems preventing this from being immediately implemented)

whatever changes i bring over, ill do my best to give attribution :)

PR #203, opened 2025-01-25
feat: kitchen sink 2.0

lxsmnsyc:
image

- Provides the barebones kitchen-sink app for react-scan. Just add a new example under `kitchen-sink/src/examples`
- Removed the old `examples` and `kitchen-sink`


kitchen-sink/src/examples/sierpinski/index.tsx
```diff
@@ -1,22 +1,17 @@
...
+// import { Analytics } from '@vercel/analytics/react';
 import { useEffect, useMemo, useState } from 'react';
-import ReactDOMClient from 'react-dom/client';
-import { scan } from 'react-scan';
+import { scan, Store } from 'react-scan';
 
 import './styles.css';
 
+
+Store.isInIframe.value = false;
```
lxsmnsyc, line 11: react-scan doesn't work inside iframes, this enables it

kitchen-sink/src/examples/todo-list/index.tsx
lxsmnsyc, line 1: I just copied this from my forgetti example: https://github.com/lxsmnsyc/forgetti/tree/main/examples/react/src

kitchen-sink/src/main.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+interface Example {
+  title: string;
+  url: string;
+}
+
+const examples: Example[] = [
+  { title: 'Sierpinski Triangle', url: '/?example=sierpinski' },
+  { title: 'Todo List', url: '/?example=todo-list' },
+];
```
lxsmnsyc, line 12: A little bit manual. When you add a new app in the `examples/`, this needs to be reflected.

`?example` query needs to match the folder name

package.json
```diff
@@ -4,7 +4,6 @@
   "scripts": {
     "build": "WORKSPACE_BUILD=true node scripts/workspace.mjs build",
     "postbuild": "node scripts/version-warning.mjs",
-    "postinstall": "pnpm build",
```
lxsmnsyc, line 7: A bit annoying.

fyi: pnpm no longer runs this by default, https://github.com/pnpm/pnpm/releases/tag/v10.0.0

pivanov (commented):
Nice one!!!

package.json
lxsmnsyc: is this part of the build?

lxsmnsyc: weird, I cannot repro this in my device, perhaps the existence of the dist folder itself. hmmm.

pivanov: oh ... yeah ... you should remove all dist folders (fresh clone)

lxsmnsyc:
long overdue

PR #215, opened 2025-01-28
perf: outlines

lxsmnsyc:


packages/scan/src/new-outlines/types.ts
```diff
@@ -10,34 +10,13 @@ export interface OutlineData {
...
-  /**
-   * height
-   */
-  number,
-  /**
-   * didCommit
-   */
-  0 | 1,
+  id: number,
+  count: number,
```
lxsmnsyc, line 14: named tuples is a TS feature, no need for comments

PR #223, opened 2025-01-31
fix: react-scan auto on playwright

lxsmnsyc:
Currently, React Scan doesn't work when there isn't a React DevTools installed (since RDT initializes the global hook React listens to **before React is loaded**), which also makes our CLI not work properly.

This fix ensures that the auto script is loaded first before React.


packages/scan/src/cli.mts
```diff
@@ -237,13 +237,14 @@ const init = async () => {
   const urlString = inferValidURL(inputUrl);
 
   await page.goto(urlString);
-  await page.waitForLoadState('load');
-  await page.waitForTimeout(500);
 
-  await page.addScriptTag({
+  await page.addInitScript({
```
lxsmnsyc, line 241: This runs before any script in the initial html, unlike `addScriptTag` which runs after the page is loaded.

packages/scan/src/cli.mts
```diff
@@ -298,7 +299,7 @@ const init = async () => {
 
       await page.evaluate(() => {
         if (typeof globalThis.reactScan !== 'function') return;
-        globalThis.reactScan({ report: true });
+        globalThis.reactScan({});
```
lxsmnsyc, line 302: deprecated config.

packages/scan/src/core/index.ts
```diff
@@ -69,12 +56,15 @@ const initRootContainer = (): RootContainer => {
...
   fragment.appendChild(iconSprite);
   fragment.appendChild(cssStyles);
   shadowRoot.appendChild(fragment);
 
-  document.documentElement.appendChild(rootContainer);
+  const currentContainer = rootContainer;
+
+  requestAnimationFrame(() => {
+    document.documentElement.appendChild(currentContainer);
+  });
```
lxsmnsyc, line 67: the script executed by Playwright runs such that `document.documentElement` is null. By using RAF we defer appending the root container.

packages/scan/src/new-outlines/index.ts
```diff
@@ -523,15 +516,18 @@ export const initReactScanInstrumentation = () => {
...
-      const host = getCanvasEl();
-      if (host) {
-        document.documentElement.appendChild(host);
-      }
-      globalThis.__REACT_SCAN__ = {
-        ReactScanInternals,
-      };
-      startReportInterval();
-      logIntro();
+      requestAnimationFrame(() => {
```
lxsmnsyc, line 519: Same as the previous one, we want to defer this one to ensure that the documentElement exists.

RobPruzan:
I'm not seeing the scans on the following:

```bun run packages/scan/dist/cli.js https://tracing.cse.buffalo.edu/student/tracing/practice```

image

lxsmnsyc:
deprecated

PR #228, opened 2025-02-02
fix: fix and improve timeline updates and fiber tracking

pivanov:
- Optimize fiber tree traversal by caching root fiber
- Fix timeline batch updates
- Improve props/state change tracking logic
- Fix wrapper types count display


RobPruzan (approved):
LGTM

PR #231, opened 2025-02-03
perf: toolbar

lxsmnsyc:
- This PR migrates the current toolbar code into fine-grained rendering/reactivity. We are already utilizing Preact Signals, let's use it to the max.


lxsmnsyc:
redoing

PR #232, opened 2025-02-04
feat: pannels

pivanov:


packages/scan/src/web/views/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { Header } from '~web/widget/header';
+import { ViewInspector } from './inspector';
+import { ViewSlowDowns } from './slow-downs';
+import { Toolbar } from './toolbar';
+
+export const Content = () => {
+  const refContainer = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    const unsubscribeStoreInspectState = Store.inspectState.subscribe(
```
lxsmnsyc, line 15: We could perhaps use `useSignalEffect`, just to simplfy stuff (One thing I learned during the rewrite)

RobPruzan: oh cool, does it automatically track dependencies?

pivanov: yeah we can I just try to do it one way because of the perf optimisations it's easy to find all subscriptions and refactor them at once :) we have places with few subscriptions and I just don't wanna mess up diff implementations ... but you are absolutely right!!! :)

PR #238, opened 2025-02-07
fix: playwright and react 18 compat

lxsmnsyc:
- Fixes script injection in playwright, ensures React Scan is evaluated first before any script.
- Fixes toolbar initialization to only happen after React finishes hydration (not compatible pre-React 19)


packages/scan/src/core/index.ts
```diff
@@ -497,8 +496,9 @@ export const start = () => {
 
   const options = getOptions();
 
-  idempotent_createToolbar(!!options.value.showToolbar);
-  initReactScanInstrumentation();
+  initReactScanInstrumentation(() =>
```
lxsmnsyc, line 499: Callback magic

packages/scan/src/core/instrumentation.ts
```diff
@@ -338,6 +338,7 @@ interface InstrumentationConfig {
   onCommitFinish: OnCommitFinishHandler;
   onError: OnErrorHandler;
   onActive?: OnActiveHandler;
+  onPostCommitFiberRoot: () => void;
```
lxsmnsyc, line 341: We are hooking into React's schedule after effects has finished.

packages/scan/src/new-outlines/index.ts
```diff
@@ -480,6 +478,13 @@ export const initReactScanInstrumentation = () => {
     onCommitFinish: () => {
       ReactScanInternals.options.value.onCommitFinish?.();
     },
+    onPostCommitFiberRoot() {
+      const host = getCanvasEl();
+      if (host) {
+        document.documentElement.appendChild(host);
+      }
+      setupToolbar();
```
lxsmnsyc, line 486: Only create the toolbar and the canvas after commit.

This needs debate since we are basically only showing the toolbar after the initial render/hydration.

packages/scan/src/cli.mts
```diff
@@ -237,12 +237,12 @@ const init = async () => {
   const urlString = inferValidURL(inputUrl);
 
   await page.goto(urlString);
-  await page.waitForLoadState('load');
-  await page.waitForTimeout(500);
 
-  await page.addScriptTag({
```
RobPruzan, line 243: did this not have a guarantee our script ran first?

packages/scan/src/new-outlines/index.ts
RobPruzan: This is definitely fine in terms of UX. Do we know why this is needed specifically when running in playwright? Or is this a workaround of a bippy bug

packages/scan/src/core/instrumentation.ts
RobPruzan: oh wait is this a guarantee? Thinking back to our conversation in slack about tracking effect profiling, I guess onCommitFiberRoot -> onPostCommitFiberRoot would encapsulate that (but ofc not precise)

packages/scan/src/new-outlines/index.ts
lxsmnsyc: addInitScript runs before the documentElement even exists, and onActive runs during that phase (not sure why)

RobPruzan: oh that's weird

packages/scan/src/cli.mts
lxsmnsyc: no, this runs after all script has been evaluated

packages/scan/src/core/instrumentation.ts
lxsmnsyc: oCFR seems to happen before effects are done. I'm not exactly sure if that's the correct timing. I decided to got with oPCFR based on the timing.

PR #241, opened 2025-02-10
fix: remove unused refs

lxsmnsyc:
So I found a handful of it, and they decided to remove them. These are basically memory leaks

There are also refs that can be removed if we changed on derived attributes/styles rather than updating these imperatively, but I'll probably change them later


packages/scan/src/web/views/inspector/what-changed.tsx
```diff
@@ -932,6 +931,7 @@ const CountBadge = ({
     }
 
     if (forceFlash) {
+      // is ref necessary
```
pivanov, line 934: this one I think was because of fast updates but we can make it without refs

packages/scan/src/web/views/toolbar/index.tsx
```diff
@@ -13,7 +13,7 @@ import { constant } from '~web/utils/preact/constant';
 import FpsMeter from '~web/widget/fps-meter';
 
 export const Toolbar = constant(() => {
-  const refSettingsButton = useRef<HTMLButtonElement>(null);
+  // const refSettingsButton = useRef<HTMLButtonElement>(null);
```
pivanov, line 16: I think we can clean this one ... having settings.tsx (commented) is enough I think

pivanov (approved):
Nice!!! ❤️

packages/scan/src/web/views/toolbar/index.tsx
lxsmnsyc: well the only reason I didn't remove it is the huge chunk of comments below it. I'll probably remove this if you want the rest removed as well

PR #247, opened 2025-02-12
fix: useSignalEffect instead of subscribe

lxsmnsyc:
Continuation of #241 but this time it focuses on cleaning up some rendering logic involving imperative DOM manipulation into declarative form.

bonus: simplified conditional class names

### DIFF
--- a/packages/scan/src/web/views/index.tsx
+++ b/packages/scan/src/web/views/index.tsx
@@ -1,32 +1,17 @@
+import { type ReadonlySignal, useComputed } from '@preact/signals';
 import type { ReactNode } from 'preact/compat';
-import { useEffect, useRef } from 'preact/hooks';
 import { Store } from '~core/index';
 import { signalWidgetViews } from '~web/state';
-import { cn, toggleMultipleClasses } from '~web/utils/helpers';
+import { cn } from '~web/utils/helpers';
 import { Header } from '~web/widget/header';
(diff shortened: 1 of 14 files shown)


packages/scan/src/web/views/index.tsx
```diff
@@ -44,15 +29,17 @@ export const Content = () => {
...
+        className={useComputed(() =>
+          cn(
+            'relative',
+            'flex-1',
+            'flex flex-col',
+            'rounded-t-lg',
+            'overflow-hidden',
+            'opacity-100',
+            'transition-[opacity]',
+            isInspecting.value && 'opacity-0 duration-0 delay-0',
```
pivanov, line 41:
```
{
  'opacity-0 duration-0 delay-0': isInspecting.value,
}
```
this one is a bit better I think because:

- It avoids unnecessary class strings when isInspecting.value is false (i.e., no extra "false" or empty classes).
- Tailwind’s JIT mode can optimize it better since it directly maps classes to conditions.

pivanov (approved):
Nice! ❤️

packages/scan/src/web/views/index.tsx
lxsmnsyc: > no extra "false" or empty classes
oh wait, `cn` behaves like that for booleans?

PR #250, opened 2025-02-13
fix: toolbar schedule

lxsmnsyc:
in #238 , we moved the toolbar and outline initialization into `onPostCommitFiberRoot`, however, the problem with that hook is that it only runs when there's at least 1 `useEffect` or one passive action (e.g. unmounting element) queued in React's workloop.

This fix ensures that the toolbar runs at least 1 second after the last render triggered.

### DIFF
--- a/packages/scan/src/new-outlines/index.ts
+++ b/packages/scan/src/new-outlines/index.ts
@@ -433,6 +433,26 @@ export const isValidFiber = (fiber: Fiber) => {
 export const initReactScanInstrumentation = (setupToolbar: () => void) => {
   if (hasStopped()) return;
   // todo: don't hardcode string getting weird ref error in iife when using process.env
+  let schedule: ReturnType<typeof setTimeout>;
+  let mounted = false;
+
+  const scheduleSetup = () => {
+    if (mounted) {
+      return;
+    }
+    if (schedule) {
+      clearTimeout(schedule);
+    }
+    schedule = setTimeout(() => {
+      mounted = true;
+      const host = getCanvasEl();
+      if (host) {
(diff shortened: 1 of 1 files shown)


pivanov (approved):

pivanov:
@lxsmnsyc 
Lemme explain what I've done in 314ce17c47574e21cf98db62dfb691468fb988ba

**Problem:**
Previously, we used page.addScriptTag() to inject React Scan, but this led to race conditions with React's initialization. The script would load after React had already started, causing us to miss initial events and requiring manual initialization and page reloads as workarounds.

**Solution:**
- Replace page.addScriptTag with context.addInitScript for reliable initialization
- Remove redundant script injection and manual reactScan({}) call
- Add early return after reload to prevent double initialization

Implementation:
Old (**addScriptTag**):
1. Page loads and begins executing JavaScript
2. React initializes and sets up fiber tree
3. React Scan script loads via addScriptTag (too late - React already initialized)
4. Manual reactScan({}) call needed to attempt initialization
5. Additional reload required to actually catch events

New (**addInitScript**):
1. React Scan script registered at context level (before page loads)
2. Page reloads to activate the init script
3. React Scan hooks initialize before ANY JavaScript executes
4. React initializes and sees our properly registered hooks
5. All events captured from the start, including initial render

The page reload is necessary because context.addInitScript only registers scripts
for future page loads. By reloading immediately after registration, we ensure
React Scan's hooks are properly initialized before React starts, enabling complete
event capture without manual intervention or timing-based workarounds.

P.S. I think we can make it work even if the target page already use react-scan ... just need to check few things and I will try to make it

is this make sense?

PR #256, opened 2025-02-14
feat: what changed section UI improvements

pivanov:


packages/scan/src/web/views/inspector/diff-value.tsx
```diff
@@ -1,9 +1,104 @@
...
+      type="button"
+      onClick={onToggle}
+      className="flex items-center p-0 opacity-50"
+    >
+      <Icon
+        name="icon-chevron-right"
+        size={12}
+        className={cn('transition-transform text-[#4ade80]', {
+          'rotate-90': expanded,
+          'text-[#f87171]': isNegative,
```
lxsmnsyc, line 29: we could probably shift now to conditionals. `isNegative ? 'text-[#f87171]' : 'text-[#4ade80]'` perhaps

pivanov: yeah ... fixing few more of these too :)

PR #257, opened 2025-02-17
feat: lite

lxsmnsyc:
- This PR adds a `lite` mode to react-scan. Basically it removes the toolbar and only leaves the outline overlay.
- This PR also improves the tree-shaking since our code currently has so many side-effects (fun fact: Preact is not side-effect free).
- I've also removed files we are not using since the new outline logic has been implemented.


lxsmnsyc:
Closing this

PR #258, opened 2025-02-17
fix: toolbar schedule (again)

lxsmnsyc:
1s is too long for debouncing, that React-generated animations keep deferring the schedule. This PR allows for room to breathe.

(still dies in React 18 :( )

### DIFF
--- a/packages/scan/src/new-outlines/index.ts
+++ b/packages/scan/src/new-outlines/index.ts
@@ -433,7 +433,7 @@ export const isValidFiber = (fiber: Fiber) => {
 export const initReactScanInstrumentation = (setupToolbar: () => void) => {
   if (hasStopped()) return;
   // todo: don't hardcode string getting weird ref error in iife when using process.env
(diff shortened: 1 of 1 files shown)


lxsmnsyc:
well Remix is a bit broken (mostly because of delayed hydration)

pivanov:
hmm ... something like this?
[patch.txt](https://github.com/user-attachments/files/18846997/patch.txt)

but still ... in remix we should run the rdt hook before everything else ...  because of:
image

lxsmnsyc:
> but still ... in remix we should run the rdt hook before everything else ... because of:

The thing is, most of the Remix users (and Remix itself) is in React 18, where hydration yields an error if the entire document doesn't match. React Scan inserts elements which causes the hydration to fail.

pivanov:
@lxsmnsyc yeah that's why I use `requestIdleCallback` in the patch it works on my side but still toolbar is not open because of rdt hook

lxsmnsyc:
Anyways, I'll pretend Remix doesn't exist

PR #260, opened 2025-02-17
Add Bun instructions in the README

igor9silva:
Bun is the fastest packager manager and IMHO should be the default package manager for most projects.

 For a performance tool like `react-scan`, the most performant one should be highlighted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,15 +30,15 @@ Airbnb&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;<a href="https://polaris.shopify.com/"
 ### Package managers
 
 ```bash
+bun add react-scan
+```
+
+```bash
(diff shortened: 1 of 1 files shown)


lxsmnsyc:
> Bun is the fastest packager manager and IMHO should be the default package manager for most projects.

The guide is not about which is fast, it's about which one is widely used.

The bun installation guide could've been just added, no need to push out Yarn from the guide.

igor9silva:
@lxsmnsyc I see no reason to use `yarn` in 2025, but def. no married to this idea.

lxsmnsyc:
> not married to this idea.

exactly, other people also doesn't have a reason to use `bun` either. We each have our own preferences.

igor9silva:
Added `yarn` back.

PR #273, opened 2025-02-25
feat: notifications

RobPruzan:
moved to #299


.vscode/settings.json
```diff
@@ -7,7 +7,7 @@
   },
   "css.lint.unknownAtRules": "ignore",
   "[typescript]": {
-    "editor.defaultFormatter": "biomejs.biome"
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
```
lxsmnsyc, line 10: oh boy

RobPruzan: don't worry i turned linting off entirely :)

packages/scan/src/core/heatmap-overlay.ts
```diff
@@ -0,0 +1,237 @@
...
+    highlightCanvas.style.height = `${rect.height}px`;
+    highlightCanvas.width = rect.width * dpr;
+    highlightCanvas.height = rect.height * dpr;
+    highlightCtx.scale(dpr, dpr);
+
+    drawHighlights();
+  };
+  handleResizeListener = handleResize;
+
+  window.addEventListener("resize", handleResize);
```
lxsmnsyc, line 217: no debounce/throttle?

packages/scan/src/core/index.ts
```diff
@@ -497,16 +529,56 @@ export const start = () => {
 
   const options = getOptions();
 
-  idempotent_createToolbar(!!options.value.showToolbar);
-  initReactScanInstrumentation();
 
-  const isUsedInBrowserExtension = typeof window !== 'undefined';
+  initReactScanInstrumentation({
+    onActive: () => {
+      const rdtHook = getRDTHook();
```
lxsmnsyc, line 535: if my assumption here is correct (I haven't seen the implementation), this might not be compatible with React 18 sync SSR (not Remix-based since they cheat). Basically, the issue is that React 18 SSR is strict with the markup during hydration so any tampering would cause it throw error. React 19 doesn't have this issue thanks to the new SSR architecture.

packages/scan/src/core/notifications/event-tracking.ts
```diff
@@ -0,0 +1,418 @@
...
+  createHighlightCanvas(document.body);
+
+  const onComplete = async (
+    _: string,
+    finalInteraction: FinalInteraction,
+    event: PerformanceEntryChannelEvent
+  ) => {
+    toolbarEventStore.getState().actions.addEvent({
+      kind: "interaction",
+      id: crypto.randomUUID(),
```
lxsmnsyc, line 365: future fix: needs compat check

packages/scan/src/core/notifications/performance-utils.ts
```diff
@@ -0,0 +1,114 @@
...
+      queue.push([child, node]);
+    }
+  }
+  return tree;
+};
+
+const isProduction: boolean = process.env.NODE_ENV === "production";
+const prefix: string = "Invariant failed";
+
+// FIX ME THIS IS PRODUCTION INVARIANT LOL
```
lxsmnsyc, line 66: it's fine, I do this as well LOL

packages/scan/src/core/notifications/performance.ts
```diff
@@ -0,0 +1,1110 @@
...
+const DEFAULT_FILTERS: PathFilters = {
+  skipProviders: true,
+  skipHocs: true,
+  skipContainers: true,
+  skipMinified: true,
+  skipUtilities: true,
+  skipBoundaries: true,
+};
+
+const FILTER_PATTERNS = {
```
lxsmnsyc, line 48: realistically this may not be 100% (nor 50%) accurate, unless this is intentional, of course.

@popover.tsx
lxsmnsyc, line 1: empty file ;)

packages/scan/src/core/index.ts
lxsmnsyc: This also throws an error for the CLI since the auto script runs before the document exists.

packages/scan/src/core/notifications/event-tracking.ts
RobPruzan: oops missed this, should use the generateId fn we have

packages/scan/src/core/notifications/performance.ts
RobPruzan: oh yeah that was from monitoring a while ago. I think i had o1 think of a while and create a ton of test cases that filtered bad components, and that was the implementation that passed the most cases. Not really necessary to do this for notifications

packages/scan/src/core/index.ts
RobPruzan: do u know how we solved this on current main?

lxsmnsyc: @RobPruzan https://github.com/aidenybai/react-scan/blob/97b9de66180dce4449fe938cfaf60a9e32eb499b/packages/scan/src/new-outlines/index.ts#L439-L454

debounce until no commit is done to the fiber tree, which is basically two birds hit with one stone.

PR #302, opened 2025-03-01
feat: extension v1.0.6

pivanov:


packages/extension/src/content/index.ts
```diff
@@ -22,60 +21,31 @@ chrome.runtime.onMessage.addListener(
...
-        showToolbar: isEnabled,
-      });
 
-      broadcast.onmessage = (type, data) => {
-        if (type === 'react-scan:react-version' && data.version) {
-          sendResponse({ hasReact: toggledState });
-        }
-      };
 
-      broadcast.postMessage('react-scan:toggle-state', { state: toggledState });
```
RobPruzan, line 62: what does "background" mean in this context

packages/extension/src/inject/index.ts
```diff
@@ -1,72 +1,154 @@
...
+      },
+    },
+  );
 
-    reactScan.scan(options);
-  },
-);
+  if (toggledState) {
+    // we need to reload because of the instrumentation
+    window.location.reload();
```
RobPruzan, line 87: is this a special case? The page didn't reload when u screenshared (unless this is for something else)

packages/extension/src/background/icon.ts
```diff
@@ -1,217 +1,50 @@
...
+  badgeText?: string,
+): Promise<void> => {
   try {
-    if (enabled) {
-      if (isFirefox) {
-        // Use animated SVG for Firefox
+    switch (state) {
+      case IconState.ENABLED:
         await browserAction.setIcon({
+          tabId: tab.id,
```
RobPruzan, line 19: what happens if we don't scope it to a tab. Does this mean u can enable on one tab, and disable on another? That's nice if so

pivanov: yeap it works per tab

packages/extension/src/inject/index.ts
pivanov: yes ... for some reason (need to check it because the icon issues are related to this) when we remove the toolbar and create it again we loose the  `fiberRoots`

packages/extension/src/content/index.ts
pivanov: mmm  not sure which background do you mean? In extension context background usually means background script or worker (depends on the manifest version)

RobPruzan: Yeah, assume I know nothing about this. What context this snippet running in:
>  broadcast.postMessage('react-scan:toggle-state', { state: toggledState });

is this the same thread as the users website? Is this a worker and you're posting to another worker/background script (and what is the relevant distinction between a background and a worker)

pivanov: oh broadcast is the Broadcast API (native browser api) who send messages to all tabs and windows from same origin.

The background script/worker works in extension context ... it's a bit complex at first look but it's not that much when you get the concept

pivanov:
Close this one in favor of #315

PR #307, opened 2025-03-03
docs: vite plugin in vite.md

ramu-narasinga:
This PR has the following changes:

1. Vite Plugin TODO is updated with an example snippet from packages/vite-plugin-react-scan, with a note pointing to README in packages/vite-plugin-react-scan

### DIFF
--- a/docs/installation/vite.md
+++ b/docs/installation/vite.md
@@ -37,7 +37,26 @@ scan({
 
 ## Vite plugin
 
-TODO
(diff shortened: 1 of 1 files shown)


ramu-narasinga:
Hey @aidenybai, not sure what `## Preserving component names` mean in this same file, vite.md. Could you please point me in the right direction or provide more context so I could also complete that TODO? thanks.

PR #328, opened 2025-03-13
feat: show rendered parents in notifications view

RobPruzan:
Recursively shows the parents (that rendered) of notification render bars

This is a little tricky to implement perfectly (w.r.t UX) since it's trivial to create a cycle:
```html
<Div>
  <Button>
    <Div>
      <Button/>
    </Div>
  </Button>
</Div>
```

In this case Div is a parent of Button, and vice versa. So this tree would recurse forever in notifications. 

This a limit to how the fiber renders datastructure is aggregated. We group fibers that are created from the same component together, and just store the count. So we have no way to express how many layers this cyclical structure holds. We can only track that some Div's had a Button parent, and some Buttons had a Div parent

This actually is fine in most cases, and I could not think of a trivial way to represent this better, while allowing memory to scale O(1) w.r.t grouped fibers

Demo: https://x.com/milliondotjs/status/1900041985378914464

### DIFF
--- a/packages/scan/src/web/views/notifications/render-bar-chart.tsx
+++ b/packages/scan/src/web/views/notifications/render-bar-chart.tsx
@@ -4,6 +4,7 @@ import { getIsProduction } from '~core/index';
 import { iife } from '~core/notifications/performance-utils';
 import { cn } from '~web/utils/helpers';
 import {
+  GroupedFiberRender,
   NotificationEvent,
   getTotalTime,
   isRenderMemoizable,
@@ -13,6 +14,7 @@ import {
   HighlightStore,
   drawHighlights,
 } from '~core/notifications/outline-overlay';
+import { ChevronRight } from './icons';
 
 // todo: cleanup, convoluted ternaries
 export const fadeOutHighlights = () => {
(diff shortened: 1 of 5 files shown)


packages/scan/src/web/views/notifications/render-bar-chart.tsx
```diff
@@ -48,20 +50,21 @@ export const fadeOutHighlights = () => {
   };
```
RobPruzan, line 50: Not really a lot of changes below, looks a lot worse because of a code move (extracting inline JSX to a component)

packages/scan/src/web/views/notifications/render-bar-chart.tsx
```diff
@@ -154,275 +152,368 @@ export const RenderBarChart = ({
...
+  };
+  totalBarTime: number;
+  isProduction: boolean | null;
+}) => {
+  const { setNotificationState, setRoute } = useNotificationsContext();
+  const [isExpanded, setIsExpanded] = useState(false);
+
+  const isLeaf = bar.kind === 'render' ? bar.event.parents.size === 0 : true;
+
+  const parentBars = bars.filter((otherBar) =>
```
RobPruzan, line 206: if the parentBar is a render, get the associated render bars to recursively render

packages/scan/src/web/views/notifications/render-bar-chart.tsx
```diff
@@ -154,275 +152,368 @@ export const RenderBarChart = ({
...
+    otherBar.kind === 'render' && bar.kind === 'render'
+      ? bar.event.parents.has(otherBar.event.name) &&
+        otherBar.event.name !== bar.event.name
+      : false,
+  );
+
+  const missingParentNames =
```
RobPruzan, line 213: It will always be the case (unless the entire tree renders) that a fiber that rendered will have a parent that didn't render. 

This is the "leaf node" to our bottom up tree, but i'd still like to visualize this because if the parent that didn't render is a user component, and the rendered component is a node_module, it's high signal to know where the node_module is rendered

packages/scan/src/web/views/notifications/render-bar-chart.tsx
```diff
@@ -154,275 +152,368 @@ export const RenderBarChart = ({
...
+              'absolute right-0 top-1/2 transition-none -translate-y-1/2 bg-white text-black px-2 py-1 rounded text-xs opacity-0 group-hover:opacity-100 transition-opacity mr-16',
+              'pointer-events-none',
+            ])}
+          >
+            Click to learn more
+          </div>
+        )}
+      </div>
+
+      {isExpanded &&
```
RobPruzan, line 485: the recursive render, not much happening here

packages/scan/src/web/views/notifications/render-bar-chart.tsx
```diff
@@ -154,275 +152,368 @@ export const RenderBarChart = ({
...
+            bar.kind === 'render' && !isLeaf && setIsExpanded(!isExpanded)
+          }
+          className={cn([
+            'flex items-center min-w-fit shrink-0 hover:bg-[#0f0f0f] rounded-r-md h-[28px]',
+            bar.kind === 'render' && !isLeaf
+              ? 'cursor-pointer'
+              : 'cursor-default',
+          ])}
+        >
+          <div className="w-[20px] flex items-center justify-center">
```
RobPruzan, line 447: idk im not in love with the UI but it does the trick

PR #339, opened 2025-03-17
fix(vite-plugin-react-scan): replace isProduction with isBuild

### LINKED ISSUE #338: [vite-plugin-react-scan] Enabling react-scan in vite build with NODE_ENV=development doesn’t work
When running vite build with NODE_ENV=development, react-scan does not get enabled as expected. The issue stems from checking config.isProduction https://github.com/aidenybai/react-scan/blob/main/packages/vite-plugin-react-scan/src/index.ts#L192

It would be more appropriate to check config.command === 'serve' to determine when and how to include react-scan

This affects cases where we want to enable react-scan during development-like builds but outside of the typical Vite serve command.

piecyk:
fixes #338

### DIFF
--- a/packages/vite-plugin-react-scan/src/index.ts
+++ b/packages/vite-plugin-react-scan/src/index.ts
@@ -92,7 +92,7 @@ const reactScanPlugin = (options: ReactScanPluginOptions = {}): Plugin => {
   } = options;
 
   let config: ResolvedConfig;
-  let isProduction = false;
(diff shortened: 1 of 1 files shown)


pivanov:
@piecyk please check my comment https://github.com/aidenybai/react-scan/issues/338#issuecomment-2730536410

pivanov (approved):
r+

PR #342, opened 2025-03-18
fix(vite-plugin-react-scan): recursively resolve react-scan content in monorepo

### LINKED ISSUE #341: [vite-plugin-react-scan] Fix module resolution for hoisted react-scan in monorepo setup
In a monorepo setup, the react-scan module is sometimes hoisted to the root node_modules, causing module resolution issues. The current implementation assumes that react-scan exists locally in the module's node_modules, leading to errors when it's hoisted.

Proposed Fix:
Implement a recursive lookup strategy to resolve the correct path to react-scan and read its content.

piecyk:
fixes #341

### DIFF
--- a/packages/vite-plugin-react-scan/src/index.ts
+++ b/packages/vite-plugin-react-scan/src/index.ts
@@ -6,6 +6,21 @@ import * as cheerio from 'cheerio';
 import type { Options } from 'react-scan';
 import type { Plugin, ResolvedConfig } from 'vite';
 
+async function resolveModuleFileContent(moduleName: string, startDir: string = process.cwd()) {
+  const modulePath = path.join('node_modules', moduleName);
+  const resolvedPath = path.resolve(startDir, modulePath);
+
+  try {
+    return await fs.promises.readFile(resolvedPath, 'utf-8');
+  } catch {
+    const parentDir = path.dirname(startDir);
+    if (parentDir !== startDir) {
+      return resolveModuleFileContent(moduleName, parentDir);
+    }
+    throw new Error(`Module ${moduleName} not found`);
+  }
+}
+
 interface Logger {
   debug: (...args: unknown[]) => void;
   info: (...args: unknown[]) => void;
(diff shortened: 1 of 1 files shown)


packages/vite-plugin-react-scan/src/index.ts
```diff
@@ -6,6 +6,21 @@ import * as cheerio from 'cheerio';
...
+async function resolveModuleFileContent(moduleName: string, startDir: string = process.cwd()) {
+  const modulePath = path.join('node_modules', moduleName);
+  const resolvedPath = path.resolve(startDir, modulePath);
+
+  try {
+    return await fs.promises.readFile(resolvedPath, 'utf-8');
+  } catch {
+    const parentDir = path.dirname(startDir);
+    if (parentDir !== startDir) {
+      return resolveModuleFileContent(moduleName, parentDir);
```
hshoja, line 18: Maybe stops the upward traversal if found the monorepo root (package.json) as well

piecyk: Good idea. On the other hand, in most cases, this plugin requires react-scan to be installed, so adding the stop might introduce unnecessary complexity. Let's see what @pivanov thinks.

pivanov: Thank you @hshoja for bringing up this consideration about monorepo handling!
I have a few thoughts to share:

Since `react-scan` is explicitly defined as a required (non-optional) peer dependency in our package.json:
```json
"peerDependenciesMeta": {
  "react-scan": {
    "optional": false
  }
}
```

1. The current directory traversal behavior actually serves a useful purpose:
   - It handles both standard and monorepo setups transparently
   - It works correctly with hoisted dependencies, which is a common pattern in monorepos
   - It follows the Node.js module resolution algorithm that users expect

2. Adding a stop condition for monorepo roots could potentially:
   - Break valid setups where dependencies are hoisted to a higher level
   - As @piecyk said ... will add complexity to handle edge cases (what defines a monorepo root? package.json with workspaces? pnpm-workspace.yaml?)
   - Create confusion when the dependency is actually available but we stop looking for it

For these reasons, I believe we should keep the current implementation as it provides the most reliable and predictable behavior across different repository structures. The focus should remain on ensuring `react-scan` is properly installed as a peer dependency.

piecyk: Thanks for the detailed explanation, @pivanov! Ping me if anything is needed to merge this.

pivanov: No additional changes are needed :) gimme few seconds

pivanov:
@piecyk v1.0.8 is available :) Thank you!

PR #354, opened 2025-04-02
fix: update fiber renderData for both function and object types

jazelly:
This fixes a bad logic in the `updateFiberRenderData` util.


aidenybai:
ty!

PR #360, opened 2025-04-09
chore: early return traversal when unnecessary render found

jazelly:


jazelly:
After playing with this more, realized this is wrong. One prop unchanged doesn't mean anything, and we have to traverse all the props.

PR #362, opened 2025-04-12
chore: early return traversal when unnecessary render found

jazelly:
For a rendered fiber, if one prop change found necessary, we can say the fiber render is necessary, and no need to check the rest props. This PR hanldes that so it early returns `traverseProps`


arkhurst:
➕

PR #366, opened 2025-04-17
fix: add missing title attribute

johanwulf:
Certain buttons that only has icons is missing the title attribute. This causes accessibility scanners, for instance Axe, to fail when having react-scan active on a webpage.

image

This title adds the missing attributes that I could identify. If there are more, I'm happy to fix those as well.

### DIFF
--- a/packages/scan/src/web/views/toolbar/index.tsx
+++ b/packages/scan/src/web/views/toolbar/index.tsx
@@ -151,6 +151,7 @@ export const Toolbar = constant(() => {
         <button
           type="button"
           id="react-scan-notifications"
+          title="Notifications"
           onClick={() => {
             if (Store.inspectState.value.kind !== 'inspect-off') {
               Store.inspectState.value = {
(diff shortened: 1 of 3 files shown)


johanwulf:
I ran formatting, but it seems to do changes in unrelated files. Formatting doesn't seem to have been ran on the main branch.

packages/scan/src/web/views/toolbar/index.tsx
```diff
@@ -207,6 +208,7 @@ export const Toolbar = constant(() => {
         checked={!ReactScanInternals.instrumentation?.isPaused.value}
         onChange={onToggleActive}
         className="place-self-center"
+        title="Visualise component updates"
```
RobPruzan, line 211: Can we change this to "Outline Re-renders"?

RobPruzan (changes requested):

packages/scan/src/web/views/toolbar/index.tsx
johanwulf: Done

RobPruzan:
Thanks for the PR!

Issue #356, opened 2025-04-04
number of renders shows for all instances


codemonkeynorth:
If i click a button in the top toolbar, the number of renders is updating agaisnt both instances. Shouldn't each Toolbar show its renders separately?

https://codesandbox.io/p/devbox/react-scan-test-forked-rwm7jx?file=%2Fsrc%2FApp.tsx&workspaceId=ws_S6K2DvAuMPjqyGLwMX3Ruk

demo: https://rwm7jx.csb.app/

![Image](https://github.com/user-attachments/assets/450a3803-df13-4521-a1c6-b0c7e4aee119)

thanks
J

tengweiherr:
Hi React Scan team,

I took a look into this issue out of curiosity, I hope my findings and suggestions are helpful!

**[Root cause]**
I believe this is due to [renderDataMap](https://github.com/aidenybai/react-scan/blob/main/packages/scan/src/core/instrumentation.ts#L441) which I believe is used to track renders in the component tree. In [instrumentation.ts](https://github.com/aidenybai/react-scan/blob/main/packages/scan/src/core/instrumentation.ts#L441), `renderDataMap` is implemented as a WeakMap, and we are using the fiber's `type` (function / class associated to the fiber) as the key. 

In the example above, it doesn't work because both the fibers share the same `type`. It means they share the same key in the `renderDataMap`, hence both have the same renderCount. This issue also persists even if we remove the keys:
```
  <Toolbar toolbar={topToolbar} />
  <Toolbar toolbar={bottomToolbar} />
```

**[Suggestions]**
My proposed solution is quite straightforward:
- incorporate additional fiber attributes like `fiber.key` and `fiber.index` into the key.
- at the same time, ensure the keys remain stable across renders — meaning we can't use the fiber object as a key, as its reference changes in every React commit

My idea is to convert `renderDataMap` from a `WeakMap` to `Map`
```
export const renderDataMap = new Map<string, RenderData>();
```
and construct a stable composite key derived from the fiber:
```
export function getRenderDataMapKey(fiber: Fiber) {
  const type = String(fiber?.type);
  return `${type}-${fiber?.key ?? ''}-${fiber?.index ?? ''}`;
}

const renderKey = getRenderDataMapKey(fiber)

renderDataMap.set(renderKey, { ...renderData });
```
This allows us to identify the correct fiber using a combination of type, key, and index, therefore accurate render tracking

[...]

PR #369, opened 2025-04-23
fix: duplicate render count when one component is reused multiple times

tengweiherr:
fix https://github.com/aidenybai/react-scan/issues/356

I was waiting for feedback from the team, but I think it's better to open a PR so everyone can have a look into the proposed solution :) Suggestions are welcome!

**[Root cause]**
I believe this is due to [renderDataMap](https://github.com/aidenybai/react-scan/blob/main/packages/scan/src/core/instrumentation.ts#L441) using the fiber's `type` (function / class associated to the fiber) as the key.
In the example from the issue, it doesn't work because both the fibers share the same `type` reference. Since they share the same key in the `renderDataMap`, hence both have the same value (renderCount). 
This issue also happens without the keys:
```
  <Toolbar toolbar={topToolbar} />
  <Toolbar toolbar={bottomToolbar} />
```

**[Proposed solution]**
My proposed solution is quite straightforward:
- Use a composite key with additional fiber attributes like `fiber.key` and `fiber.index`
- Instead of storing a `RenderData` object as the value, store a Map with the composite key and the `RenderData` as the value

This allows us to cross check with the key and index to identify the correct fiber

**[Result]**

https://github.com/user-attachments/assets/54fc64b1-e379-4671-94d2-52151a3ff9cd

### DIFF
--- a/packages/scan/src/core/instrumentation.ts
+++ b/packages/scan/src/core/instrumentation.ts
@@ -438,17 +438,25 @@ export interface OldRenderData {
 
 const RENDER_DEBOUNCE_MS = 16;
 
-export const renderDataMap = new WeakMap<object, RenderData>();
+export const renderDataMap = new Map<string, RenderData>();
+
+export function getRenderDataMapKey(fiber: Fiber) {
+  const typeName = String(fiber?.type);
+  return `${typeName}-${fiber?.key ?? ''}-${fiber?.index ?? ''}`;
+}
 
 const trackRender = (
-  type: unknown,
+  fiber: Fiber,
   fiberSelfTime: number,
   fiberTotalTime: number,
   hasChanges: boolean,
   hasDomMutations: boolean,
 ) => {
   const currentTimestamp = Date.now();
-  const existingData = renderDataMap.get(type as object);
+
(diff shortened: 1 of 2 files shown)


tengweiherr:
Before (recorded on the [demo link](https://rwm7jx.csb.app/))

https://github.com/user-attachments/assets/350f67fa-d119-49a3-a4f9-842a7e7ac482

tengweiherr:
**[Before]**
(recorded on the [demo link](https://rwm7jx.csb.app/))



https://github.com/user-attachments/assets/8f39a47e-b9c6-4d43-ad72-08dc821c0326

RobPruzan:
Will give this a look soon, pls spam me if I don't review in the next couple days

tengweiherr:
@RobPruzan 

Tested my PR in a more complex React app and found a bug. I fixed it by changing the data structure to a WeakMap that holds a nested Map, allowing it to hold additional fiber attributes while tying its lifetime to the fiber's reference

Feel free to test it on your end as well :)

tengweiherr:
bumping this again :P @RobPruzan

RobPruzan:
LGTM

Issue #355, opened 2025-04-04
not compatible with @emotion/styled


codemonkeynorth:
if the returned component  starts with eg a `styled.div` react-scan does not update the render count and does not show the main component name
https://codesandbox.io/p/sandbox/wys52t?file=%2Fsrc%2Fcomponents%2FToolbar.tsx%3A3%2C8

```js
const Toolbar = ({ toolbar }) => {
  // ... etc
  
 return (
    <StyledToolbar>
    some content
   </StyledToolbar>
  )
}

const StyledToolbar = styled.div``
```

open the preview to see react-scan working
https://wys52t.csb.app/

![Image](https://github.com/user-attachments/assets/36fb6d87-d0bb-4878-bc53-d1e76c3c2b0f)

thanks
J

lxsmnsyc:
> does not show the main component name

You're required to use `displayName` for styled components.

codemonkeynorth:
@lxsmnsyc sorry I forgot to add that that's an option, but then the render counts still don't show 

And actually it should theoretically show 2 components? The Toolbar and the StyledToolbar (which might rerender due to its own props?)

tengweiherr:
Some preliminary findings

I think the issue might be related to `forwardRef` — react scan's inspector doesn’t seem to pick up components wrapped with `forwardRef`.

I took a look at [how Emotion generates components](https://github.com/emotion-js/emotion/blob/cce67ec6b2fc94261028b4f4778aae8c3d6c5fd6/packages/styled/src/base.tsx#L122-L188), and they’re [wrapping components with forwardRef under the hood](https://github.com/emotion-js/emotion/blob/cce67ec6b2fc94261028b4f4778aae8c3d6c5fd6/packages/react/src/context.tsx#L30-L47). So I tried swapping the Emotion styled component in the example with a plain forwardRef component like this:
```
const ForwardRefToolbar = forwardRef((props, ref) => {
  return <div>{props.children}</div>;
});
```

Same result — not picked up by the inspector. 

I guess it might not work for styled-components either, based on [how components are generated](https://github.com/styled-components/styled-components/blob/76b18a4cb94da357cafa8c0d9bc1e96372fee980/packages/styled-components/src/models/StyledComponent.ts#L264).

But do note that in React 19, [`forwardRef` is no longer needed](https://react.dev/reference/react/forwardRef) and will be deprecated in future releases. Passing ref as a prop should work fine with the current react scan.

tengweiherr:
[...]

PR #376, opened 2025-05-11
fix: compatible with styled components

tengweiherr:
fix https://github.com/aidenybai/react-scan/issues/355

`@emotion/styled` and `styled-components` [call components with forwardRef](https://github.com/emotion-js/emotion/blob/cce67ec6b2fc94261028b4f4778aae8c3d6c5fd6/packages/react/src/context.tsx#L30-L47) when generating them. We weren’t retrieving the fiber type from ForwardRef correctly, as it has a different data structure compared to functional and host components

To fix it, we just need to make a one line change: use `getRenderData(getType(node?.fiber))` instead of `getRenderData(node?.fiber.type)`.

Also, I cleaned up the code a bit as I just found out that Bippy already provides [getFiberId](https://github.com/aidenybai/bippy/blob/main/packages/bippy/src/core.ts#L555-L575) to manage uniqueId for fibers, so we don't really need to implement our own fiber identifier.

Tested on the preview link https://wys52t.csb.app mentioned in https://github.com/aidenybai/react-scan/issues/355

https://github.com/user-attachments/assets/914b3334-35ce-423e-9e29-c08a2dd4c2dc

### DIFF
--- a/packages/scan/src/core/instrumentation.ts
+++ b/packages/scan/src/core/instrumentation.ts
@@ -10,6 +10,7 @@ import {
   SimpleMemoComponentTag,
   didFiberCommit,
   getDisplayName,
+  getFiberId,
   getMutatedHostFibers,
   getTimings,
   getType,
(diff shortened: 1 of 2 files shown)


RobPruzan:
LGTM, nice work 🚀

PR #382, opened 2025-05-21
Update BROWSER_EXTENSION_GUIDE.md with Chrome web store link

stanhuan:
Noticed that there is an official Chrome extension so updating the documentation to include a link to it. Also moved some other content around


RobPruzan:
Awesome, thanks!

PR #396, opened 2025-06-29
fix: prevent window dragging in Electron apps by adding WebkitAppRegion no-drag

jvnkr:
## Description

This PR adds the `WebkitAppRegion: 'no-drag'` CSS property to the React Scan widget to prevent accidental window dragging when users interact with the toolbar in Electron applications.

## Changes

- Added `style={{ WebkitAppRegion: 'no-drag' }}` to the main widget container div
- This prevents the Electron window from being dragged when users interact with React Scan controls

## Why this change?

In Electron applications, by default, users can drag the window by clicking and dragging on any part of the interface. This can interfere with the React Scan widget's functionality, as users might accidentally drag the window when trying to interact with the toolbar buttons or resize handles.

The `WebkitAppRegion: 'no-drag'` property specifically tells Electron that this area should not trigger window dragging behavior.


RobPruzan:
nice! thanks :)

PR #397, opened 2025-07-02
fix: scan does not accept expected options

torytang025:
**Title:**  
`scan` does not accept expected options such as `_debug`, `onCommitStart`, `onRender`, `onCommitFinish`, `onPaintStart`, `onPaintFinish`

```ts
scan({
  enabled: true,
  showFPS: true,
  showToolbar: true,

 // the following attributes are ignored
  _debug: "verbose",
  onRender: () => {},
  onCommitStart: () => {},
  onCommitFinish: () => {},
  onPaintStart: () => {},
  onPaintFinish: () => {},
});
```

**Description**

When passing the following options to `scan`, they are not recognized as expected:

- `_debug`
- `onCommitStart`
- `onRender`
- `onCommitFinish`
- `onPaintStart`
- `onPaintFinish`

**Investigation**

After some investigation, I found that in `packages/scan/src/core/index.ts`, the function for filtering option keys is as follows:

```ts
function isOptionKey(key: string): key is keyof Options {
  return key in ReactScanInternals.options.value;
}
```

However, the default options defined in `ReactScanInternals.options` do **not** include the above parameters, so they are filtered out during this check.

```ts
export const ReactScanInternals: Internals = {
  instrumentation: null,
  componentAllowList: null,
  options: signal({
    enabled: true,
    // includeChildren: true,
    // playSound: false,
    log: false,
    showToolbar: true,
    // renderCountThreshold: 0,
    // report: undefined,
    // alwaysShowLabels: false,
    animationSpeed: 'fast',
    dangerouslyForceRunInProduction: false,
    showFPS: true,
    showNotificationCount: true,
    allowInIframe: false,
    // smoothlyAnimateOutlines: true,
    // trackUnnecessaryRenders: false,
  }),
  runInAllEnvironments: false,
  onRender: null,
  scheduledOutlines: new Map(),
  activeOutlines: new Map(),
  Store,
  version: packageJson.version,
};
```

**Fix Description**
This fix refactors how option keys are filtered and stored in `scan`:

1. **Removal of `isOptionKey`:**  
   I believe the `isOptionKey` function is unnecessary for the following reasons:
   - Any unexpected properties will not be assigned to `validOptions` during validation, so they are naturally filtered out.
   - The function is heavily dependent on the default options. If new default options are added in the future, they could easily be missed, leading to maintenance issues.

2. **Improved Local Storage Filtering:**  
   When saving options to local storage, the code now uses the `LocalStorageOptions` type to filter out unnecessary parameters, such as callback functions and internal flags. This ensures that only relevant options are persisted, and special parameters like `onCommitStart`, `onRender`, etc., are excluded from storage but still available for use at runtime.


pivanov (approved):
will r+ this one but I think @RobPruzan can merge it

packages/scan/src/core/index.ts
```diff
@@ -342,18 +342,24 @@ export type LocalStorageOptions = Omit<
   | 'onPaintFinish'
 >;
 
-function isOptionKey(key: string): key is keyof Options {
-  return key in ReactScanInternals.options.value;
-}
+const filetLocalStorageOptions = (options: Options): LocalStorageOptions => {
```
RobPruzan, line 345:
```suggestion
const applyLocalStorageOptions = (options: Options): LocalStorageOptions => {
```

RobPruzan:
thanks for the contribution!

torytang025:
@RobPruzan Hi! I noticed that after this PR was merged, the function filetLocalStorageOptions was removed and replaced with applyLocalStorageOptions by commit [589b2dc](https://github.com/aidenybai/react-scan/pull/397/commits/589b2dc863c8a5f945eb2e295bfc6a484ce3f8d5), but the call site wasn't updated accordingly, which is now causing a runtime error.

I've submitted a follow-up PR to fix it: [#398]. Thanks!

RobPruzan:
> @RobPruzan Hi! I noticed that after this PR was merged, the function filetLocalStorageOptions was removed and replaced with applyLocalStorageOptions by commit [589b2dc](https://github.com/aidenybai/react-scan/pull/397/commits/589b2dc863c8a5f945eb2e295bfc6a484ce3f8d5), but the call site wasn't updated accordingly, which is now causing a runtime error.
> 
> I've submitted a follow-up PR to fix it: [#398]. Thanks!

oops 😭 , thanks for the catch

